PR #58, opened 2023-10-26
Fix #48 (libbpf building issue)

### LINKED ISSUE #48: [BUG] ci does not use project libbpf
**Describe the bug**

0766a16ff1b686470f7ecf9bdaa468b

**To Reproduce**

make unit-test

it seems that the program is using local headers

Officeyutong:
This PR downgrades libbpf to a certain version that can be both run on Debian 12 and Ubuntu 22.

It also fixed some issues when creating customize commands to the copy_headers target, so that no system-wide headers will be used

Note: I chose a version of libbpf that does not provide functions like `bpf_{prog,map}_get_info_by_fd`, use `bpf_obj_get_info_by_fd`, they are the same things 

Closes #48


runtime/src/bpf_map/array_map_kernel_user.cpp
```diff
@@ -34,7 +34,7 @@ void array_map_kernel_user_impl::init_map_fd()
 	}
 	bpf_map_info info = {};
 	unsigned int info_len = sizeof(info);
-	int res = bpf_map_get_info_by_fd(map_fd, &info, &info_len);
```
yunwei37, line 37: Why?

Officeyutong: I chose a version of libbpf that does not provide functions like bpf_{prog,map}_get_info_by_fd, use bpf_obj_get_info_by_fd, they are the same things

yunwei37 (approved):

PR #95, opened 2023-11-20
third_party: add catch2 as submodule

yunwei37:
## Description

Add catch2 as submodule, and remove some useless code.

This may avoid the download process at build time, make it suitable for publish on other platforms.


Officeyutong:
verifier(A submodule) uses Catch2 using FetchContent

yunwei37:
Yes, I noticed that...if we needs to get this merge, maybe we need to pr the upstream repo.

PR #120, opened 2023-12-03
runtime: remove attach type replace, rename filter to override

### LINKED ISSUE #112: [FEATURE] Remove deprecated attach type replace
bpf_override_return helper can be used for replacing the replace/filter types. we needs to remove the deprecated code and fix the test cases.

Officeyutong:
This PR removes the implementation of attach type replace in the attach manager, and modifies tests that shipped with replace to override. 

Replace and filter were unified to override now. The override callback was performed before the hooked function's execution. It has access to the arguments, and could override the return value of the hooked function using `bpftime_set_retval`. If `bpftime_set_retval` was called, the original function will not be executed. Otherwise, the execution of the hooked function was not affected.

Closes #112

### DIFF
--- a/runtime/src/attach/attach_manager/frida_attach_manager.cpp
+++ b/runtime/src/attach/attach_manager/frida_attach_manager.cpp
@@ -84,7 +84,7 @@ int frida_attach_manager::attach_at(void *func_addr, callback_variant &&cb)
 			      .first;
 		SPDLOG_DEBUG("Created frida attach entry for func addr {:x}",
 			     (uintptr_t)func_addr);
-	} else if (itr->second->has_replace_or_override()) {
+	} else if (itr->second->has_override()) {
 		SPDLOG_ERROR(
 			"Function {} was already attached with replace or filter, cannot attach anything else");
 		return -EEXIST;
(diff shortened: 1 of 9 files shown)


runtime/src/attach/attach_manager/frida_attach_manager.cpp
```diff
@@ -368,21 +373,21 @@ void frida_internal_attach_entry::iterate_uretprobe_callbacks(
...
-	convert_gum_cpu_context_to_pt_regs(*ctx->cpu_context, regs);
-	frida_internal_attach_entry *hook_entry =
-		(frida_internal_attach_entry *)
-			gum_invocation_context_get_replacement_data(ctx);
-	auto &cb = hook_entry->get_replace_callback();
-	uint64_t ret = cb(regs);
-	gum_invocation_context_replace_return_value(ctx, (gpointer)ret);
-	return ret;
-}
+// extern "C" uint64_t __bpftime_frida_attach_manager__replace_handler()
```
yunwei37, line 376: Remove them.

Comment out code is unnecessary and will bring more maintaining problems, so please don't do that.

yunwei37 (changes requested):

runtime/src/attach/attach_manager/frida_attach_manager.cpp
Officeyutong: Fixed now

yunwei37 (approved):

PR #131, opened 2024-01-04
vm/runtime: implement userspace -> kernel tailcall

Officeyutong:
Implement userspace to kernel tailcall.

Userspace ebpf programs might utilize bpf_tail_call helper and BPF_MAP_TYPE_PROG_ARRAY. fds in BPF_MAP_TYPE_PROG_ARRAY should be kernel prog fds. 

kernel programs to call could be arbitary types that bpf_prog_test_run supports (see https://docs.kernel.org/bpf/bpf_prog_run.html#id1 for full list). In our unit tests, we use BPF_PROG_TYPE_RAW_TRACEPOINT.

Context passed through bpf_tail_call should either be a null pointer (where bpftime will provide 64-byte context to the kernel program), or an at-least 64 byte buffer.


vm/simple-jit/ebpf_vm.c
```diff
@@ -989,14 +994,16 @@ int ebpf_exec(const struct ebpf_vm *vm, void *mem, size_t mem_len,
 					__ATOMIC_RELAXED);
 				break;
 			}
-			// case EBPF_CMPXCHG: {
```
yunwei37, line 992: why commented out these code?

yunwei37 (approved):

yunwei37:
Would be better if you can have a README.md for the tail call example.

vm/simple-jit/ebpf_vm.c
Officeyutong: They were introduced in a previous commit, I'll check it

Officeyutong:
> Would be better if you can have a README.md for the tail call example.

I'll add it later

PR #139, opened 2024-01-14
runtime: implement userspace USDT

### LINKED ISSUE #135: [FEATURE] USDT support
In addition to current u(ret)probe support, it would be useful to support USDT.

Support libbpf and bpftrace's [`usdt`](https://github.com/iovisor/bpftrace/blob/master/docs/reference_guide.md#9-usdt-static-tracing-user-level)

Not quite sure. Maybe the current uprobe support can support usdt as well?

**Provide usage examples**

Trace usdt (such as the DTrace probes in OpenJDK https://github.com/openjdk/jdk/blob/e44276989fc6358065412be7567d0141c84f1282/src/hotspot/os/posix/dtrace/hotspot.d#L4)

Officeyutong:
Closes #135 

> On x86_64, a usdt tracepoint was corresponded to a nop instruction in the code. So we may hook this instruction to implement userspace USDT I've tested with GumInvocationListener of frida gum, it works well when attached to the address of the nop instruction.
> 
> So things we need to add USDT support are:
> 
> * logic to parse ELF and extract USDT definitions
> 
> Other logics would be similar to uprobe, and can even be reused

### DIFF
--- a/.github/workflows/test-examples.yml
+++ b/.github/workflows/test-examples.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Install dependencies
         run: | 
           sudo apt-get update
-          sudo apt-get install binutils-dev libboost1.74-all-dev libelf-dev zlib1g-dev libyaml-cpp-dev gcc-12 g++-12 llvm systemtap-sdt-dev
+          sudo apt-get install binutils-dev libboost1.74-all-dev libelf-dev zlib1g-dev libyaml-cpp-dev gcc-12 g++-12 llvm
       - name: Build and install runtime (with llvm-jit)
         if: ${{matrix.enable_jit}}
         run: |
@@ -146,7 +146,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y zlib1g-dev libelf-dev llvm
+          sudo apt-get install -y zlib1g-dev libelf-dev llvm systemtap-sdt-dev
           sudo mkdir /mnt/ramdisk
       - name: Build test assets
         run: |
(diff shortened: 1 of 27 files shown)


Officeyutong:
According to libbpf's description to usdt https://github.com/libbpf/libbpf/blob/f81eef23b33c0dbf923e863a72ce51ea4d32e291/src/usdt.c#L24

The implementation of attach type usdt is just plain uprobes, attached to the specified nop instruction. libbpf handles all other things, including parse ELF and figure out tracepoints, resolving tracepoint arguments in the ebpf side, and so on, through two internal maps (__bpf_usdt_specs and __bpf_usdt_ip_to_spec_id, where the latter is only needed if bpf cookie is not supported)

So the thing we need to do is 
- Implement bpf cookie
- Test directly with usdt programs, and check if it can work. If it can't work, just fix bugs. We should have already supported USDT

caizixian:
This looks great! Thanks!

Would you mind also checking whether this works with bpftrace (I remember seeing somewhere that you can use bpftrace with bpftime as the runtime)?

Officeyutong:
> This looks great! Thanks!
> 
> Would you mind also checking whether this works with bpftrace (I remember seeing somewhere that you can use bpftrace with bpftime as the runtime)?

Ok, I'll also check it

Officeyutong:
USDT is almost working now. Bpftrace is pending for testing

yunwei37:
Maybe we can get this PR merged first if the libbpf example in `example/usdt_minimal/README.md` is working? 

After that, we can:
1. check and fix bpftrace
2. update documents.

Officeyutong:
> Maybe we can get this PR merged first if the libbpf example in `example/usdt_minimal/README.md` is working?
> 
> After that, we can:
> 
> 1. check and fix bpftrace
> 2. update documents.

It's OK to merge now

yunwei37 (approved):

PR #147, opened 2024-01-20
example: add a uprobe example to trace go programs

Officeyutong:
WIP. Expected to be done in these two days

### DIFF
--- a/example/minimal-go-uprobe/.gitignore
+++ b/example/minimal-go-uprobe/.gitignore
@@ -1,0 +1,12 @@
+.vscode
+package.json
+*.o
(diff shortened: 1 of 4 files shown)


yunwei37:
Could you please add the example to https://github.com/eunomia-bpf/bpf-developer-tutorial/tree/main/src/31-goroutine ?

I think there is no need to put it here.

Officeyutong:
> Could you please add the example to https://github.com/eunomia-bpf/bpf-developer-tutorial/tree/main/src/31-goroutine ?
> 
> I think there is no need to put it here.

OKay

PR #201, opened 2024-01-29
runtime: Fix runtime epoll_wait error when timeout is set to -1

### LINKED ISSUE #186: [BUG] Hot looping in bpftime's epoll_wait emulation
I've found that bpftime's epoll_wait emulation might hot loop when no events arrive. The `top` shows the tracer process (or the bpftime server process's child process) is almost always using 100% CPU:

```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
 644087 agentzh   20   0   32640  10028   9600 R 100.0   0.0   1:50.25 ebpf_loader
```

`perf top` shows the hottest functions are all in `libbpftime-syscall-server.so`:

```
+   68.52%     0.52%  libbpftime-syscall-server.so  [.] ring_buffer__poll
+   67.28%     1.55%  libbpftime-syscall-server.so  [.] epoll_wait
+   60.51%     1.15%  libbpftime-syscall-server.so  [.] syscall_context::handle_epoll_wait
+   35.74%     5.84%  libc-2.31.so                  [.] __GI___sigprocmask
```

Can we add some sleep in between? This is too wasteful when the events do not arrive immediately.

Zheaoli:
## Description

Fixes #186

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Firmware version:
- Hardware:
- Toolchain:
- SDK:

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

### DIFF
--- a/runtime/src/bpftime_shm.cpp
+++ b/runtime/src/bpftime_shm.cpp
@@ -322,7 +322,7 @@ int bpftime_epoll_wait(int fd, struct epoll_event *out_evts, int max_evt,
 		auto now_time = high_resolution_clock::now();
 		auto elasped =
 			duration_cast<milliseconds>(now_time - start_time);
-		if (elasped.count() > timeout) {
+		if (timeout && elasped.count() > timeout) {
 			break;
 		}
 		for (const auto &p : epoll_inst.files) {


Zheaoli:
https://elixir.bootlin.com/linux/v6.5-rc2/source/fs/eventpoll.c#L1843

Like what we have already done in Linux kernel , we need to check the timeout value in epoll_wait emulation

BTW, I prefer to replace all the timer from loop check to `select` event based, WDYT @yunwei37 ?

yunwei37:
Thank you very much for your help!

> BTW, I prefer to replace all the timer from loop check to select event based, WDYT

However, one problem may be that, if we want to have a better performance, we should avoid execute any syscall in the helpers of eBPF runtime. execute syscalls in the helpers will greatly slow down the runtime.

The ring buffer helpers work similar to iouring, maybe iouring can help, but if we rely on a feature only supports by higher version of Linux, the portable advantage of bpftime will be lost.

yunwei37:
I think it's a bad idea to mock epoll syscalls in bpftime, but I didn't find a better solution... maybe you can help us?

The epoll is used to notify the eBPF applications (eg. bpftrace) there are elements available in the ring buffer.

Maybe we can register some socket fd (make libbpf assume they are ring buffer maps fd) to epoll, and use them to notify the eBPF applications? 

The notification may only happens if the ring buffer is full, so not every bpf helper call will bring a syscall.

yunwei37:
Or maybe we can simply set a flag in the shared memory to tell the eBPF runtime the ring buffer is empty, before the eBPF applications enter the select syscalls. So the eBPF runtime in another process can filled the selected fd.

yunwei37 (approved):

PR #208, opened 2024-01-31
fix newline and add gitattributes

### LINKED ISSUE #183: [BUG] Found many text files like CMakeLists.txt use the MS-DOS newline sequence (CR LF)
As title. We should always use the UNIX newline?

yunwei37:
## Description

Fixes #183, change some file endline from CRLF to LF.

Add a `.gitattribute` like:

```c
* text eol=lf
```

May avoid future code using CRLF.

## Type of change


Officeyutong:
Wow, so many lines!

Officeyutong (approved):

PR #209, opened 2024-01-31
build in parallel in makefile

### LINKED ISSUE #173: [BUG] There is no easy way to build in parallel on SMP systems
Currently `make ... -jN` does not work since the Makefile invokes `cmake --build` instead of `$(MAKE)` so flags like `-jN` won't get passed down to the sub-make which does the real work.

Maybe we can apply a patch like this?

https://gist.github.com/agentzh/6ef6ee15d5e131675dd2992289fc860e

Then we can invoke it on an SMP Linux system like this:

```bash
make ... JOBS=$(nproc)
```

Without parallelism, it is too slow to build bpftime.

yunwei37:
## Description

Fixes #173

The build time is:

```console
$ time make build JOBS=8
real    1m47.211s
user    5m14.661s
sys     0m45.855s
```

which is similar to:

```console
$ cmake -Bbuild -DBPFTIME_ENABLE_UNIT_TESTING=1 -DBUILD_BPFTIME_DAEMON=1 -DCMAKE_BUILD_TYPE:STRING=Debug
$ time cmake --build build --config Debug -j8
real    1m43.163s
user    5m13.930s
sys     0m36.892s
```

Documents updated in https://eunomia.dev/bpftime/documents/build-and-test/#compilation-for-bpftime

## Type of change

- [X] This change requires a documentation update


Officeyutong:
Would -j$(nproc) be better?

yunwei37:
Seems there are no easy ways for makefile to get -j number.

> In a Makefile, the -j option is used with the make command to specify the number of jobs (commands) to run simultaneously. However, within the Makefile itself, you cannot directly retrieve the number passed to the -j option because it's an argument to the make command and not an environment variable or a Makefile variable.
>
> If you want to act on the -j option within a Makefile, you typically have to pass it as a variable. However, GNU make provides a way to detect if the job server is enabled (which happens when you use -j without specifying the number of jobs) via the .NOTPARALLEL special target. But it does not provide a way to get the actual number of jobs specified by the user.

From GPT4

Officeyutong (approved):

PR #215, opened 2024-02-03
test: use Catch2 and cmake for runtime tests

Officeyutong:
This PR uses cmake and catch2 to organize all actively-maintained runtime tests:
- Build all unit tests into a single binary
- eBPF programs that these tests need would be compiled by CMake in a proper order
- unit test programs will be able to find the ebpf programs automatically

So after this PR, we only need the following command to run runtime tests:
```bash
cmake -DBPFTIME_LLVM_JIT=YES -DBPFTIME_ENABLE_UNIT_TESTING=YES -DCMAKE_BUILD_TYPE=Release -B build
cmake --build build --config Release --target bpftime_runtime_tests
sudo ./build/runtime/unit-test/bpftime_runtime_tests
```

This PR also updates `test-runtime.yml` to adapt the new tests


Officeyutong:
> Are the deleted old unit tests already covered by the new unit tests?
> 
> If not, I think before this pr gets merged, we need to add more tests to ensure the coverage.
> 
> You can add new unit tests in another PR?

I've covered all old unit tests that were enabled in runtime/test/CMakeLists.txt

Other remained test sources seem to be obsoleted. They were not listed in the CMakeLists, and would not be executed when running make unit-test. And some of them were not enable to run under the current bpftime(used outdated functions). I kept them here for future purpose(e.g migrate those tests to the new unit test framework)

.github/workflows/test-runtime.yml
```diff
@@ -1,8 +1,9 @@
 name: Build and test runtime
 
 on:
+  workflow_dispatch: 
   push:
-    branches: "master"
```
yunwei37, line 5: why change here?

yunwei37 (approved):

.github/workflows/test-runtime.yml
Officeyutong: Pushes to branches other than main should also trigger the CI, I think. This would be convenient for non-pull request debugging purpose

Officeyutong: It's also convenient for consuming CI time of my own account instead of the organization account. In this way I don't need a pull request to test the workflows in my own fork for branch other than main

PR #218, opened 2024-02-12
attach: reconstruct attach related parts

Officeyutong:
This PR reconstructs the whole attach parts of bpftime/runtime. They are under `/attach`

There are four major targets:
- `/attach/base_attach_impl`: Provide an abstraction of all attach impls. Any attach impl which implements this base class could be registeres to bpftime runtime (when other parts of bpftime/runtime were reconstructed), without modifying bpftime itself
- `/attach/frida_uprobe_attach_impl`: Uprobe implementation based on frida. Migrated from the old attach manager. 
- `/attach/syscall_trace_attach_impl`: Syscall trace implementation based on zpoline. Split out from the old bpf_attach_ctx
- `/attach/text_segment_transformer`: moved from the old `agent-transformer`

Also, tests were made for `syscall_trace_attach_impl` and `frida_uprobe_attach_impl`

Docs are in header files.

Other changes were made to bpftime:
- Almost everything under `/runtime/src/attach` was removed. The implementation of attaching was moved out
- Tests related that doesn't require ebpf programs were moved to `/attach`

- [x] tests
- [x] integrate with old code base
- [x] syscall trace parts
- [x] full docs
- [x] text segment transformer - integrate to new attach folder


CMakeLists.txt
```diff
@@ -98,6 +101,7 @@ add_subdirectory(tools)
 # benchmark that requires bpftime libraries
 add_subdirectory(benchmark)
 
+
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
 
-install(TARGETS bpftime-agent bpftime-agent-transformer bpftime-syscall-server CONFIGURATIONS Release Debug RelWithDebInfo DESTINATION ~/.bpftime)
+install(TARGETS bpftime-agent bpftime_text_segment_transformer bpftime-syscall-server CONFIGURATIONS Release Debug RelWithDebInfo DESTINATION ~/.bpftime)
```
yunwei37, line 107: is it possible not change the install name?

Officeyutong: This is the target name. The file name was kept the same as before, see https://github.com/Officeyutong/bpftime/blob/8526a42652bca7ecadd9df5525003c4307a5ebfa/attach/text_segment_transformer/CMakeLists.txt#L6

yunwei37 (approved):

PR #220, opened 2024-02-18
ci: Release binary with RelWithDebInfo and support "no-omit-frame-pointer" to keep the frame pointer

### LINKED ISSUE #174: [BUG] We should include debuginfo even in the release builds otherwise it's hard to debug
I suggest we change the `Release` cmake build type occurrences in the Makefile to `WithDebugInfo` instead. It's better to have DWARF debug info even for the release builds. Otherwise, it's really hard to debug things.

Zheaoli:
Fix #174 add more info to help the debug action


Officeyutong:
Please also update https://github.com/eunomia-bpf/eunomia.dev/blob/main/docs/bpftime/documents/build-and-test.md

Issue #155, opened 2024-01-23
[FEATURE] Implement a better AOT cache 


yunwei37:
**Is your feature request related to a problem? Please describe.**

The current AOT cache is not complete, and it should be implemented and control by the bpftime runtime. There may also be some security issues with that.

**Describe the solution you'd like**

I think maybe we should remove the current AOT cache in llvm jit, and design a better implementation for that.

PR #227, opened 2024-02-25
vm: remove aot cache for security issues

yunwei37:
## Description

Remove AOT file cache for security issues. We may need a better design for it.

Related to #155

## Type of change


Officeyutong:
What are the security issues?

yunwei37:
> What are the security issues?

The AOT code is store in files and load without verify. If attacker modify the AOT code, then they can do anything in the runtime.

Officeyutong:
> > What are the security issues?
> 
> The AOT code is store in files and load without verify. If attacker modify the AOT code, then they can do anything in the runtime.

Seems the security level is the same as modifying things in memory

Officeyutong:
Did you remove the AOT cli?

PR #238, opened 2024-02-27
attach: fix the performance issue of the new attach implementation

### LINKED ISSUE #223: [BUG] Performance issue of the new attach implementation
After the new attach implementation, uprobe seems to be a bit slower (About 5%) than before

### Before
before

### After
after

Try findout the rootcause and fixit.

Some possible rootcause
- One more layer callback, see https://github.com/eunomia-bpf/bpftime/blob/236c1042c3c51ce1e476a109ab5f5b5b53fdea81/attach/frida_uprobe_attach_impl/src/frida_uprobe_attach_impl.cpp#L156

Officeyutong:
Closes #223 

This PR allows `frida_attach_entry` accepting ebpf-arguments-like callbacks, in this way we could avoid an extra callback layer that translates ebpf arguments from original arguments.

## Benchmark result of the current commit
image
## Benchmark result of the commit prior to the new attach implementation
image

According to the benchmark, the new attach implementation has the same performance as the previous one

### DIFF
--- a/attach/CMakeLists.txt
+++ b/attach/CMakeLists.txt
@@ -1,3 +1,8 @@
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+    add_definitions(-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_DEBUG)
+  else()
+    add_definitions(-DSPDLOG_ACTIVE_LEVEL=SPDLOG_LEVEL_INFO)
+  endif()
 add_subdirectory(base_attach_impl)
 add_subdirectory(frida_uprobe_attach_impl)
 add_subdirectory(syscall_trace_attach_impl)
--- a/attach/frida_uprobe_attach_impl/CMakeLists.txt
+++ b/attach/frida_uprobe_attach_impl/CMakeLists.txt
@@ -6,7 +6,7 @@ add_library(
     src/frida_attach_entry.cpp
     src/frida_attach_utils.cpp
     src/frida_attach_private_data.cpp
-    src/frida_uprobe_attach_internal.cpp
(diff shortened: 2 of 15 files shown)


yunwei37:
Can you introduce how this PR works?

Officeyutong:
> Can you introduce how this PR works? It seems like a large PR.

Allowing frida_attach_entry to store a callback function with signature int(void* memory, size_t memory_size, uint64_t* return_value). This signature is intended for calling ebpf programs. When the hook is triggered, and the mentioned callback is provided, the corresponding function would be directly triggered.

In the previous design, frida_attach_entry only accept callback with signature void(const pt_regs&). If user want to call a ebpf program with the unified registering interface, a intermediate layer that accepts pt_regs as argument and call the ebpf callback  is needed.

So the new design reduces one intermediate callback layer, so the calling latency could be reduced.

Why doesn't frida_attach_entry stores bpftime_prog to call ebpf program? Currently attach parts doesn't depend on any API in runtime, so we can't use bpftime_prog.

yunwei37:
Can we make attach part depends on the runtime API? Will that be simple?

Officeyutong:
> Can we make attach part depends on the runtime API? Will that be simple?

runtime has no need to depend on attach (except base_attach_impl), attach also has no need to depend on runtime.

They both depend on base_attach_impl. Attach impls should be registered at runtime, and attach_manager (current bpf_attach_ctx) should be responsible for managing registered attach impls. These procedure has no performance degradation. Attach callbacks are std::function which are directly provided by attach_manager. Abstraction is only needed for a unified attach impl interface

yunwei37 (approved):

PR #243, opened 2024-02-29
runtime: use bpf_link_handler to record attach links; rewrite attach part of bpf_attach_ctx

Officeyutong:
**WIP**

This PR includes:
- Rewrite `init_attach_ctx_from_handlers`, so we can use a more robust method to instantiate handlers
- Use bpf_link_handler to record all attaches. Public APIs are kept untouched

### DIFF
--- a/runtime/include/bpf_attach_ctx.hpp
+++ b/runtime/include/bpf_attach_ctx.hpp
@@ -5,6 +5,7 @@
 #include <map>
 #include <memory>
 #include <frida_uprobe_attach_impl.hpp>
+#include <set>
 #include <syscall_trace_attach_impl.hpp>
 typedef struct _GumInterceptor GumInterceptor;
 typedef struct _GumInvocationListener GumInvocationListener;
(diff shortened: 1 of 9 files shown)


runtime/CMakeLists.txt
```diff
@@ -117,24 +115,21 @@ target_include_directories(${PROJECT_NAME}
...
-  ${SYSCALL_TRACE_ATTACH_IMPL_INCLUDE}
 )
 
 target_link_libraries(${PROJECT_NAME}
   PUBLIC
   vm-bpf
   spdlog::spdlog
-  bpftime_frida_uprobe_attach_impl
-  bpftime_syscall_trace_attach_impl
+  bpftime_base_attach_impl
```
yunwei37, line 124: Is it possible to remove this line? Seems base_attach_impl is a header only library.

yunwei37 (approved):

runtime/CMakeLists.txt
Officeyutong: CMake use this way to use header only library

Issue #239, opened 2024-02-28
[FEATURE] Support bpftime uprobe on kernel that doesn't support uprobe


Officeyutong:
Libbpf will try to read /sys/bus/event_source/devices/uprobe/type when creating uprobe perf event. This file doesn't exist on kernel that doesn't support uprobe. But since userspace uprobe doesn't rely on kernel features, we should make it support such kernel versions.

Possible ways:
- Hook function open and read, so that make libbpf think the file exists

yunwei37:
In docker, you can mount a regular dir into `/sys/bus/event_source/devices/uprobe/`
![fc5ea63646aace06117ae7acfe40a45](https://github.com/eunomia-bpf/bpftime/assets/34985212/6f78fb54-36a4-4ed4-87f3-ae9dfd56baf9)

So in order to run bpftime on old kernel version, you can simply copy the content of  `/sys/bus/event_source/devices/uprobe/` on new kernel, put it on old kernel, and mount it with docker.

yunwei37:
Another approach is hook open function with LD_PRELOAD in syscall-server.so, and prepare a directory with the same content as `/sys/bus/event_source/devices/uprobe` in `~/.bpftime/event_source/devices/uprobe`. When the libbpf tries to read something in `/sys/bus/event_source/devices/uprobe` but it's not exists, we can change the open file to make it actually read contents in  `~/.bpftime/event_source/devices/uprobe`.

yunwei37:
Related to #145

yunwei37:
Please modify code in https://github.com/eunomia-bpf/bpftime/tree/master/runtime/syscall-server

zhangzihengya:
hi @Officeyutong  @yunwei37  would like to take a crack at this, could you please assign this to me

Officeyutong:
> Hi [@Officeyutong](https://github.com/Officeyutong) [@zhangzihengya](https://github.com/zhangzihengya) Could you help confirm is it fixed?

Seems it was fixed in https://github.com/eunomia-bpf/bpftime/pull/333, but this issue is not closed

yunwei37:
Thanks!

PR #248, opened 2024-03-09
Support read /sys/bus/event_source/devices/uprobe/type on old kernel versions

zhangzihengya:
solve #239


Officeyutong:
/sys/bus/event_source/devices/uprobe/type was already read by libbpf, do you have any method to solve this?

zhangzihengya:
> /sys/bus/event_source/devices/uprobe/type was also accessed by libbpf, do you have any method to solve this?

zhangzihengya:
> /sys/bus/event_source/devices/uprobe/type was also accessed by libbpf, do you have any method to solve this?

Since we can't modify libbpf directly, what I modify is the contents of runtime/syscall-server library. ld_preload can override the open function in libc to achieve the effect of modification at runtime, so as to solve this problem.

runtime/syscall-server/syscall_server_utils.cpp
```diff
@@ -94,9 +94,46 @@ static int parse_uint_from_file(const char *file, const char *fmt)
...
+            char *expanded_path = (char *)malloc(len + 1);     // +1 for the null terminator
+
+            if (expanded_path != NULL) {
+                // Construct the expanded path
+                strcpy(expanded_path, home_dir);
+                strcat(expanded_path, path + 1);  // Skip the ~ at the beginning
+
+                return expanded_path;
+            } else {
+                fprintf(stderr, "Failed to allocate memory\n");
```
Officeyutong, line 112: Please use spdlog to print logs

Officeyutong (changes requested):

runtime/syscall-server/syscall_server_utils.cpp
```diff
@@ -94,9 +94,46 @@ static int parse_uint_from_file(const char *file, const char *fmt)
 	return ret;
 }
 
+const char *expand_tilde(const char *path) {
+    if (path[0] == '~' && (path[1] == '/' || path[1] == '\0')) {
+        const char *home_dir = getenv("HOME");
+        if (home_dir != NULL) {
+            // Allocate memory for the expanded path
+            size_t len = strlen(home_dir) + strlen(path) - 1;  // -1 to exclude ~
+            char *expanded_path = (char *)malloc(len + 1);     // +1 for the null terminator
```
Officeyutong, line 103: Is the buffer allocated here leaked?

Officeyutong:
> > /sys/bus/event_source/devices/uprobe/type was also accessed by libbpf, do you have any method to solve this?
> 
> 
> 
> Yes,Since we can't modify libbpf directly, what I modify is the contents of runtime/syscall-server library. ld_preload can override the open function in libc to achieve the effect of modification at runtime, so as to solve this problem.

I didn't see where you override the open function, will it be done in this PR?

runtime/src/bpftime_shm.cpp
```diff
@@ -137,6 +137,41 @@ void bpftime_close(int fd)
 	shm_holder.global_shared_memory.close_fd(fd);
 }
 
+const char * expand_filename(const char *filename)
+{
+	const char *home_dir = getenv("HOME");
+	if (home_dir == NULL) {
+		SPDLOG_INFO("Error: HOME environment variable is not set");
```
Officeyutong, line 144: Should it be `SPDLOG_ERROR`?

runtime/syscall-server/syscall_context.hpp
```diff
@@ -23,13 +23,17 @@ class syscall_context {
 	using epoll_ctl_fn = int (*)(int, int, int, struct epoll_event *);
 	using epoll_wait_fn = int (*)(int, struct epoll_event *, int, int);
 	using munmap_fn = int (*)(void *, size_t);
+	using fopen_fn = FILE *(*)(const char *,const char *);
+	using open_fn = int (*)(const char *, int, ...);
```
Officeyutong, line 27: Do we really need to override fopen? Doesn't fopen internally call open to open files?

zhangzihengya: In fact, the internal call of fopen is the openat system call, but when the program calls fopen, it does not trigger some of the processing done by bpftime to openat.

zhangzihengya: I did my tests on openat and didn't see that calling fopen would trigger some of the processing bpftime does for openat:

Set a breakpoint for handle_openat() as follows:

```
int syscall_context::handle_openat(int fd, const char *file, int oflag, unsigned short mode)
{
	if (!enable_mock){
		printf("1:%s\n",file);
		return orig_openat_fn(fd, file, oflag, mode);
	}
	try_startup();
	printf("2:%s\n",file);
	const char * new_file = bpftime_checkfile(file);
	printf("3:%s\n",new_file);
	return orig_openat_fn(fd, new_file, oflag, mode);
}
```

Test procedure:

```
...
	const char *filename = "/sys/bus/event_source/devices/uprobe/type";
    FILE *file = fopen(filename, "r");
...
```

Tracing the program using strace found that fopen calls the openat system call:

```
...
openat(AT_FDCWD, "/sys/bus/event_source/devices/uprobe/type", O_RDONLY) = 3
...
```

Executing the command`LD_PRELOAD=build/runtime/syscall-server/libbpftime-syscall-server.so./test`produces the following output without firing the breakpoint:

```
[2024-03-12 10:00:50.742] [info] [syscall_context.hpp:93] manager constructed
Content: 9
```

Now change the test program to one that uses the openat function directly:

```
...
    const char *directory = "/sys/bus/event_source/devices";
    const char *filename = "uprobe/type";
    int dir_fd = open(directory, O_RDONLY);
    ...
    int file_fd = openat(dir_fd, filename, O_RDONLY);
...
```

Executing the command `LD_PRELOAD=build/runtime/syscall-server/libbpftime-syscall-server.so./test` produces the following output, and you can see that the breakpoint is triggered:

```
[2024-03-12 10:03:29.609] [info] [syscall_context.hpp:93] manager constructed
[2024-03-12 10:03:29.609] [info] [syscall_server_utils.cpp:24] Initialize syscall server
[2024-03-12 10:03:29][info][26465] Global shm constructed. shm_open_type 0 for bpftime_maps_shm
[2024-03-12 10:03:29][info][26465] Enabling helper groups ufunc, kernel, shm_map by default
[2024-03-12 10:03:29][info][26465] bpftime-syscall-server started
2:uprobe/type
3:uprobe/type
Content: 9
INFO [26465]: Global shm destructed
```

Does bpftime only attach dynamic libraries to the surface of the program?

Another problem is that although this modified libbpf file access, but libbpf is actually want to get/sys/bus/event_source/devices/uprobe/type of value, and for low version of the kernel, the value is not exist, When libbpf gets the wrong value, it will also report an error (libbpf: prog 'do_count': failed to create uprobe '/lib/x86_64-linux-gnu/libc.so.6:0x9f920' perf event:  Operation not permitted), when I changed the contents of the replacement file to the correct value, but the correct value does not exist for the lower kernel version, this method does not seem feasible.

Officeyutong: > I did my tests on openat and didn't see that calling fopen would trigger some of the processing bpftime does for openat:
> 
> Set a breakpoint for handle_openat() as follows:
> 
> ```
> int syscall_context::handle_openat(int fd, const char *file, int oflag, unsigned short mode)
> {
> 	if (!enable_mock){
> 		printf("1:%s\n",file);
> 		return orig_openat_fn(fd, file, oflag, mode);
> 	}
> 	try_startup();
> 	printf("2:%s\n",file);
> 	const char * new_file = bpftime_checkfile(file);
> 	printf("3:%s\n",new_file);
> 	return orig_openat_fn(fd, new_file, oflag, mode);
> }
> ```
> 
> Test procedure:
> 
> ```
> ...
> 	const char *filename = "/sys/bus/event_source/devices/uprobe/type";
>     FILE *file = fopen(filename, "r");
> ...
> ```
> 
> Tracing the program using strace found that fopen calls the openat system call:
> 
> ```
> ...
> openat(AT_FDCWD, "/sys/bus/event_source/devices/uprobe/type", O_RDONLY) = 3
> ...
> ```
> 
> Executing the command`LD_PRELOAD=build/runtime/syscall-server/libbpftime-syscall-server.so./test`produces the following output without firing the breakpoint:
> 
> ```
> [2024-03-12 10:00:50.742] [info] [syscall_context.hpp:93] manager constructed
> Content: 9
> ```
> 
> Now change the test program to one that uses the openat function directly:
> 
> ```
> ...
>     const char *directory = "/sys/bus/event_source/devices";
>     const char *filename = "uprobe/type";
>     int dir_fd = open(directory, O_RDONLY);
>     ...
>     int file_fd = openat(dir_fd, filename, O_RDONLY);
> ...
> ```
> 
> Executing the command `LD_PRELOAD=build/runtime/syscall-server/libbpftime-syscall-server.so./test` produces the following output, and you can see that the breakpoint is triggered:
> 
> ```
> [2024-03-12 10:03:29.609] [info] [syscall_context.hpp:93] manager constructed
> [2024-03-12 10:03:29.609] [info] [syscall_server_utils.cpp:24] Initialize syscall server
> [2024-03-12 10:03:29][info][26465] Global shm constructed. shm_open_type 0 for bpftime_maps_shm
> [2024-03-12 10:03:29][info][26465] Enabling helper groups ufunc, kernel, shm_map by default
> [2024-03-12 10:03:29][info][26465] bpftime-syscall-server started
> 2:uprobe/type
> 3:uprobe/type
> Content: 9
> INFO [26465]: Global shm destructed
> ```
> 
> Does bpftime only attach dynamic libraries to the surface of the program?
> 
> Another problem is that although this modified libbpf file access, but libbpf is actually want to get/sys/bus/event_source/devices/uprobe/type of value, and for low version of the kernel, the value is not exist, When libbpf gets the wrong value, it will also report an error (libbpf: prog 'do_count': failed to create uprobe '/lib/x86_64-linux-gnu/libc.so.6:0x9f920' perf event: Operation not permitted), when I changed the contents of the replacement file to the correct value, but the correct value does not exist for the lower kernel version, this method does not seem feasible.

LD_PRELOAD can only replace functions that were dynamically linked, you may see `man ld.so` for details.

My hints for replacing statically linked functions: Use a linker script to manually override the linked function to a custom one. Is this practical?

Officeyutong:
@zhangzihengya Hi, how's going on now?

Officeyutong:
The following logic should be implemented in syscall-server (not in runtime):
- Hook `open` and `openat`, create a fake fd (using open for an temporary file) and return it for access to `/sys/bus/event_source/devices/uprobe/type`
- Hook `read`, for reading request to hooked files, writting the corresponding buffer

Officeyutong:
I don't have permission to push into `zhangzihengya:master`, so I will open a new PR

Officeyutong:
Replaced by https://github.com/eunomia-bpf/bpftime/pull/333

PR #252, opened 2024-03-19
runtime: add uprobe_multi support

### LINKED ISSUE #214: [idea]: Support uprobe_multi for bpftime

**Provide usage examples**

**Additional context**

Officeyutong:
uprobe_multi is a special type of uprobe, which allows hooking multiple userspace functions in a single attach. This PR adds uprobe_multi support for uprobe_attach_impl, the corresponding example, and update CI

Closes #214 

See https://lore.kernel.org/bpf/20230424160447.2005755-1-jolsa@kernel.org/ for details

## Our design

In the kernel bpf, the attaching of uprobe_multi were implemented through creating a bpf link, and filling all information (such as pid, function offset, attach cookies) in the link opts. You may find it in bpf_program__attach_uprobe_multi

But in bpftime, all attaches are required to have an attach target (a.k.a perf event), so for uprobe_multi that doesn't give us a perf event, we have a slightly different implementation to the kernel: When creating a bpf link, we doesn't only record configurations from bpf_link_create_opts into the handler, but also do we create some perf events for the uprobe_multi links, and record them in the handler. In this way, we tear down a uprobe_multi into several simple uprobe attach targets, and with a total bpf link. This will not affect the hooking performance.

## Main changes
- Update bpf_link_handler to make it support multiple attach targets
- Adapt for multiple attach targets in a single bpf_link


runtime/src/bpftime_shm_internal.cpp
```diff
@@ -484,8 +485,27 @@ int bpftime_shm::add_bpf_link(int fd, struct bpf_link_create_args *args)
 		errno = EBADF;
 		return -1;
 	}
-	return manager->set_handler(fd, bpftime::bpf_link_handler(*args),
-				    segment);
+	int result = manager->set_handler(
```
yunwei37, line 488: Is it better to place it somewhere else for abstraction? Shall we put the attachment related code in the runtime here?

yunwei37 (commented):
Seems we still mix attach related code in runtime?

runtime/src/bpftime_shm_internal.cpp
Officeyutong: uprobe_multi was attached in a very special way, which is different to all attaches using perf event. It fills attach targets in the bpf link, and does not use perf event or other targets(i.e target_fd would be 0 when attaching a uprobe multi). So now we don't have a suitable way to handle such abstractions. Maybe in the future when we rewritting shm client, we could split creating perf events for a uprobe multi to a individual function in the runtime

Officeyutong:
Not planned now

yunwei37:
And can we reopen this?

Officeyutong:
Moves to https://github.com/eunomia-bpf/bpftime/pull/328

PR #256, opened 2024-03-22
runtime: Add cmake options to control build runtime interact with kernel map

### LINKED ISSUE #251: [FEATURE] Add cmake options to control build runtime interact with kernel maps

See #145 and #175

Add a cmake option BPFTIME_BUILD_KERNEL_BPF, which can control whether to build the shared maps between kernel and userspace. Disable these options will not build the source code in 
https://github.com/eunomia-bpf/bpftime/tree/master/runtime/src/bpf_map/shared

hp77-creator:
## Description
Adding option to disable shared_maps in CMake

Fixes #251 

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

Checked if CI is failing or not, previously default was OFF but turned it ON and for users to disable shared_map will require them to turn it off while during ON.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/cmake/frida.cmake
+++ b/cmake/frida.cmake
@@ -26,6 +26,11 @@ elseif(${FRIDA_OS_ARCH} MATCHES "linux-arm.*")
   set(FRIDA_GUM_DEVKIT_SHA256 "6b5963eb740062aec6c22c46ec2944a68006f72d290f714fb747ffe75b448a60")
   # Frida only has armhf builds..
   set(FRIDA_OS_ARCH "linux-armhf")
+elseif(${FRIDA_OS_ARCH} MATCHES "darwin-arm.*")
+  set(FRIDA_CORE_DEVKIT_SHA256 "50aea2d5dfff000ed1f1dc1fdb2db67a02e4c4f44e782fa311f8afa31df327b6")
+  set(FRIDA_GUM_DEVKIT_SHA256 "b40c08bebd6958d41d91b7819171a457448cccad5faf417c9b4901be54b6c633")
+  # for macos-arm m* chip series 
+  set(FRIDA_OS_ARCH "macos-arm64e")
 else()
   message(FATAL_ERROR "Unsupported frida arch ${FRIDA_OS_ARCH}")
 endif()
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 CMakeCache.*
 CMakeFiles/*
 [Tt]esting/*
+cmake-build-debug
 ebpf_code.h
 vm-exten.btf
 check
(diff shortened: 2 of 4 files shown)


hp77-creator:
@yunwei37 @Officeyutong Can you please let me know if there is anything missing in this PR for #251 ?

cmake/frida.cmake
```diff
@@ -26,6 +26,11 @@ elseif(${FRIDA_OS_ARCH} MATCHES "linux-arm.*")
   set(FRIDA_GUM_DEVKIT_SHA256 "6b5963eb740062aec6c22c46ec2944a68006f72d290f714fb747ffe75b448a60")
   # Frida only has armhf builds..
   set(FRIDA_OS_ARCH "linux-armhf")
+elseif(${FRIDA_OS_ARCH} MATCHES "darwin-arm.*")
```
viniciusd, line 29: Are you getting `FRIDA_OS_ARCH` `darwin-arm64e`?

From my understanding, it is gonna be `darwin-arm64e` because this string is built from the `CMAKE_SYSTEM_PROCESSOR` variable, which is x86 by default, but it can detect arm by setting the env var [CMAKE_APPLE_SILICON_PROCESSOR](https://cmake.org/cmake/help/v3.23/variable/CMAKE_APPLE_SILICON_PROCESSOR.html), but then `CMAKE_SYSTEM_PROCESSOR` is `arm64`, not `arm64e`.

On my end, to get it compiling on Mac, I only had to set `CMAKE_APPLE_SILICON_PROCESSOR` and added the branch
```cmake
elseif(${FRIDA_OS_ARCH} MATCHES "darwin-arm64")
  set(FRIDA_CORE_DEVKIT_SHA256 "7811e516e6b7bbc0153d30095560e0b1133f154060c5542764100d3e0eb2ab2b")
  set(FRIDA_GUM_DEVKIT_SHA256 "03f6085ae5330cf38e0a498784500675fc5bd7361bb551a9097ba5fe397aceda")
else()
```
Given it doesn't seem `CMAKE_SYSTEM_PROCESSOR` ever gets the `arm64e` value

yunwei37 (approved):

cmake/frida.cmake
hp77-creator: Hi @viniciusd , arm64e was on the release link of frida-gum, I had explored both, arm64 and arm64e, and here is the difference between both: https://stackoverflow.com/a/65918986/7116645

viniciusd: Yes, it makes sense to be using the arm64e, but my point is more on the `FRIDA_OS_ARCH` value, is it ever `darwin-arm64e`? From what I could gather, it will never come with the extra `e` because the string is built from `CMAKE_SYSTEM_PROCESSOR`

PR #273, opened 2024-04-10
vm: integrate ubpf & add compat layer for different vm implementations

### LINKED ISSUE #90: [FEATURE] Remove and replace simple-jit with ubpf jit

We don't want to maintain different jit. The https://github.com/eunomia-bpf/bpftime/tree/master/vm/simple-jit was original ported from ubpf, there is no need to maintain it further.

We can use JIT directly from ubpf for resource constraints device where LLVM JIT is not suitable.

Officeyutong:
Closes #90 

This PR will replace `simple-jit` with `ubpf`, and uses a compat layer to provide unified interface for llvm-jit and ubpf. Interfaces will be in the same as the current vm

## What has been done
- `vm/llvm-jit` now comes to a new target named `bpftime_llvm_jit_vm`, which could be used and built independently
- `vm/ubpf-vm` now comes to a new target named `bpftime_ubpf_vm`, which is a wrapper on ubpf, providing unified interface to operate on ubpf. We also applies some patches to the eBPF code your load, so it would be accepted by ubpf (since ubpf lacks many things, such as lddw helpers now)
- All vms now could be accessed via a unified interface defined in target `bpftime_vm` (located at `vm/vm-core`), it provides C interface, include a series functions named `ebpf_*`. 
- `bpftime_vm_compat` is a internal header only target, and is not designed to be used by users. Only vm implementations and bpftime_vm should use this target
## How to add a vm?
- Add a new target. It should implement a subclass of `vm/compat/include/bpftime_vm_compat.hpp` (This target is a header only library, we use it to avoid circular dependency between vm implementation and bpftime_vm), and function `std::unique_ptr<bpftime_vm_impl> create_vm_instance();`
- Link your target to bpftime_vm
- Now if the user links with bpftime_vm, he will use your vm!
- You don't need to modify bpftime_vm or bpftime_vm_compat. You just need to modify CMakeLists to make bpftime_vm link on your target


yunwei37:
I think we can directly use the target vm, instead of a new target called compart layer.

runtime/object/bpf_object.cpp
```diff
@@ -125,14 +125,14 @@ bpftime_object::bpftime_object(std::string_view path)
 }
 
 // open the object elf file and load it into the context
-struct bpftime_object *bpftime_object_open(const char *obj_path)
+class bpftime_object *bpftime_object_open(const char *obj_path)
 {
-	struct bpftime_object *obj = new bpftime_object(obj_path);
+	class bpftime_object *obj = new bpftime_object(obj_path);
```
yunwei37, line 130: why not just remove the struct keyword?

vm/llvm-jit/CMakeLists.txt
```diff
@@ -56,38 +56,28 @@ else()
...
-
-target_link_libraries(vm-bpf ${LLVM_LIBS} spdlog::spdlog)
-target_include_directories(vm-bpf PUBLIC ${LLVM_INCLUDE_DIRS} ${SPDLOG_INCLUDE})
-
-bpftime_add_executable(vm-llvm-example
-    example/main.cpp
-)
+if(${BPFTIME_BUILD_STANDALONE_LLVM_VM})
+    add_subdirectory(../compat ${CMAKE_CURRENT_BINARY_DIR}/bpftime_compat)
+    add_subdirectory(../../third_party/Catch2 ${CMAKE_CURRENT_BINARY_DIR}/Catch2)
```
yunwei37, line 67: Shall we only build catch2 when test is enabled?

vm/ubpf-vm/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+add_dependencies(bpftime_ubpf_vm ubpf spdlog::spdlog bpftime_vm_compat)
+
+
+target_link_libraries(bpftime_ubpf_vm PUBLIC ${UBPF_BUILD_DIR}/lib/libubpf.a bpftime_vm_compat spdlog::spdlog)
+ExternalProject_Get_Property(ubpf SOURCE_DIR)
+target_include_directories(bpftime_ubpf_vm PRIVATE ${SOURCE_DIR}/vm/inc ${UBPF_BUILD_DIR}/vm/ ${CMAKE_CURRENT_SOURCE_DIR} ${SPDLOG_INCLUDE})
+
+
+
+# target_compile_options(bpftime_ubpf_vm PUBLIC -fsanitize=address)
```
yunwei37, line 25: remove comments if not needed

vm/vm-core/include/ebpf-vm.h
yunwei37, line 1: what about still put it into `vm/include`?

Officeyutong: It's ok but other projects under /van have their own subdirectories, will we directly put src and include under vm makes user confused?

runtime/object/bpf_object.cpp
Officeyutong: My fault. I just replaced `struct bpftime_object` to `class bpftime_object` since clang telled me I shouldn't mix the usage of `struct bpftime_object` and `class bpftime_object`

yunwei37:
The documents in https://eunomia.dev/bpftime/documents/build-and-test/#build-and-install-the-complete-runtime-in-release-modewith-llvm-jit doesn't need to be changed, right?

vm/vm-core/include/ebpf-vm.h
yunwei37: I think having an include directory is not confusing? The cpp file can be put in vm-core, that's fine. We don't need a vm/src path

Officeyutong:
> The documents in https://eunomia.dev/bpftime/documents/build-and-test/#build-and-install-the-complete-runtime-in-release-modewith-llvm-jit doesn't need to be changed, right?

A little change is needed. If user want to build vm independently, `BPFTIME_BUILD_STANDALONE_VM` should be set. If user want to build llvm-jit indenendently, `BPFTIME_BUILD_STANDALONE_LLVM_VM` should be set. These options are used for including external dependencies, such as spdlog

yunwei37:
Remember to submit a PR in the website repo please.

yunwei37 (approved):

PR #295, opened 2024-05-14
fix: bash stucks when injected by bpftime agent

### LINKED ISSUE #289: [BUG] when being attached by bpftime, bash will stuck if it spawns another bash process
## Repdocuce:
- `cd examples/libbpf-tools/bashreadline && make`
- Terminal 1. `bpftime load ./readline`
- Terminal 2. `bpftime start /bin/bash --norc`

Now everything works well. But if we run `bash` in the attached bash, it will stuck and won't handle any input, until a SIGINT (Ctrl+C) was sended

Officeyutong:
Root cause has came out!

We print logs to stdout, instead of stderr. Usually, statements in bashrc might call another program and gets its stdout for the next process. We polluted stdout of these programs, causing some scripts called by bashrc not working

Fix is also simple: print logs to stderr

## How did I find that?

On my machine, startup script of nvm triggers infinite calls to `tr`, so I tried to hooked these program, and see what's the input and output, so I found something strange:

This statement would be triggered infinite times
```
  DEFAULT_IFS=" $(nvm_echo t | command tr t \\t)
"
```
and after its run, `DEFAULT_IFS` would be
```console
INFO [1735698]: Global shm des  ruc     ed
INFO [1735697]: Global shm destructed
```

This is odd. It's expecting just a `\t`, but received many characters from the log of bpftime. These stuff were directly printed using printf, I changed them to `fprintf(stderr)`, so it was fixed

So now, bashreadline was expected to work on all machines, without any special configurations

Closes #289


Officeyutong:
My `$HOME/.bashrc` is:
```bash
# ~/.bashrc: executed by bash(1) for non-login shells.

# Note: PS1 and umask are already set in /etc/profile. You should not
# need this unless you want different defaults for root.
# PS1='${debian_chroot:+($debian_chroot)}\h:\w\$ '
# umask 022

# You may uncomment the following lines if you want `ls' to be colorized:
# export LS_OPTIONS='--color=auto'
# eval "$(dircolors)"
# alias ls='ls $LS_OPTIONS'
# alias ll='ls $LS_OPTIONS -l'
# alias l='ls $LS_OPTIONS -lA'
#
# Some more alias to avoid making mistakes:
# alias rm='rm -i'
# alias cp='cp -i'
# alias mv='mv -i'
. "$HOME/.cargo/env"
export GOROOT=/opt/go
#export PATH=/opt/node-v14.21.3-linux-x64/bin:$PATH
export PATH=$PATH:$GOROOT/bin
export PATH=$PATH:~/.bpftime
export NVM_DIR="$HOME/.nvm"
[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"  # This loads nvm
[ -s "$NVM_DIR/bash_completion" ] && \. "$NVM_DIR/bash_completion"  # This loads nvm bash_completion
export HTTP_PROXY=http://192.168.15.1:2345
export HTTPS_PROXY=http://192.168.15.1:2345
export THEOS=~/theos
export http_proxy=http://192.168.15.1:2345
export https_proxy=http://192.168.15.1:2345
```

If I comment out the line with `#This loads nvm` and `# This loads nvm bash_completion`, it will just work

yunwei37 (approved):

PR #299, opened 2024-05-26
Fix performance: use new hash map implement

yunwei37:
## Description

Fixes performance issues caused by hash map. The boost interprocess map is very slow, we replace it with a fix size, open addressing hash map in vector.

## Type of change


Officeyutong:
Will this be faster than boost?

yunwei37:
much faster

Officeyutong:
> much faster

Have you tried inserting and looking up a large number of non-random data? such as use key from 0 to 10^6 but inserted in random order? boost::interprocess is a chained hash table, when is usually much faster than open addressing

But when load factor is small, memory allocation may be too slow. we need to weigh this.

Besides, it's better if bucket size comes to a primer number when you use linear probing for open addressing hash. See https://stackoverflow.com/questions/67623801/is-it-always-necessary-to-make-hash-table-number-of-buckets-a-prime-number-for-p

Reference:
- https://stackoverflow.com/questions/2556142/chained-hash-tables-vs-open-addressed-hash-tables
- https://medium.com/@pavel.odintsov/boost-unordered-map-is-a-new-king-of-data-structures-292124d3ee2

Officeyutong:
> Reference

I think a better way is chained hash table + pre-allocated memory pool for link list nodes

yunwei37:
I think you are right in the analysis. However, I think it would be better if you can answer:

- What's the most common hash table usecases? In my case, it seems the hash table is used for configuration and summary, which only has a few elements, frequently looked up, but not frequently insert and update.
- what's the implementation algorithms of kernel hash map? Why?
- Is it better if we can provide multiple hash map implementations? It can be configured when created.

yunwei37:
> bucket size comes to a primer number
You are right, that's my mistake

Officeyutong:
> I think you are right in the analysis. However, I think it would be better if you can answer:
> 
> * What's the most common hash table usecases? In my case, it seems the hash table is used for configuration and summary, which only has a few elements, frequently looked up, but not frequently insert and update.
> * what's the implementation algorithms of kernel hash map? Why?
> * Is it better if we can provide multiple hash map implementations? It can be configured when created.

- I think hashmap are usually to store discontinuous mapping data, not limited to configuration or summary. For example, some cases we may maintain a map from `pid -> some kind of data`, while the upper limit of pid is limited by the processes count, which could be large
- libstdc++ are using chained hash table, so is kernel (See https://elixir.bootlin.com/linux/latest/source/kernel/bpf/hashtab.c#L79, the bucket is a linked list) (In almost all standard libraries of languages that I known, hashmap are implemented using chained )
- Multiple implementations is ok. But I think chained are the most suitable choice

yunwei37:
Since we don't have more people working on this, I think it would be better:

1. Add a fix_size_hash_map_impl
2. Rename the old one to var_size_hash_map_impl

And merge thi PR first.

yunwei37:
Let's try to solve as many issues as possible in this week.

PR #300, opened 2024-05-30
Update README.md

yuanrui77:
Optimized the introduction document


tools/aot/README.md
```diff
@@ -85,7 +85,7 @@ You will get a native ELF file named `do_count.o`.
 You can link it with your program and execute it:
 
 ```sh
-cd example
+cd bpftime/tools/aot/example
```
Officeyutong, line 88: This is a README file under `tools/aot`, so it's ok to remain it `cd example`? What do you think?

Officeyutong (approved):

PR #302, opened 2024-06-02
hp77 | Add Option for build without libbpf

### LINKED ISSUE #175: [FEATURE] Build options to disable libbpf and Linux kernel interoperability
Currently, the bpftime build always includes libbpf and Linux kernel interoperability, which we don't need in our use cases. It would be nice to have a build option to exclude these things to have much smaller binaries (and also smaller images in the global shm?).

I heard that it might also make it easier to port bpftime to other operating systems like macOS and *BSD (or even Windows?).

hp77-creator:
## Description

Adding an option to allow `bpftime` to built without libbpf

Fixes #175 

## Type of change

- [x] This change requires a documentation update

## How Has This Been Tested?

- [ ] Running bpftime on `MacOS`

**Test Configuration**:

- Firmware version:
- Hardware:
- Toolchain:
- SDK:

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings


hp77-creator:
TODO: had to also change path for `BOOST_ROOT` to `opt/homebrew/include` to fix `boost` import errors. Will have to mention same in the documentation.

hp77-creator:
@yunwei37 @Officeyutong right now, I am facing issue in `bpftime_shm.hpp` 
It was giving error on `sys/epoll.h` not being present on system, so I added a macro-processor check but after that there are many functions which are giving error. What to do for that? Should I also make the dependent functions enclosed in a preprocessor check?

hp77-creator:
seems like `#if __linux__` is not a good way to differentiate between systems. should I go by instruction architecture differentiation?

yunwei37:
Could you please provide some detailed error information?

Officeyutong:
> @yunwei37 @Officeyutong right now, I am facing issue in `bpftime_shm.hpp`
> It was giving error on `sys/epoll.h` not being present on system, so I added a macro-processor check but after that there are many functions which are giving error. What to do for that? Should I also make the dependent functions enclosed in a preprocessor check?

epoll are not supported on platforms except Linux. But in bpftime, we should only have used some types and structs defined in this header. You may find out these types and structs, define them in the code base of bpftime

attach/CMakeLists.txt
```diff
@@ -5,5 +5,7 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
   endif()
 add_subdirectory(base_attach_impl)
 add_subdirectory(frida_uprobe_attach_impl)
-add_subdirectory(syscall_trace_attach_impl)
-add_subdirectory(text_segment_transformer)
+if(${BPFTIME_BUILD_WITH_LIBBPF})
```
Officeyutong, line 8: Use a name like "BPFTIME_BUILD_ON_X86" or somethings alike to control these two folders would be better

Officeyutong (changes requested):

attach/CMakeLists.txt
hp77-creator: Sure will do.

hp77-creator:
getting these now:
@Officeyutong  @yunwei37 
```
bpftime-hp/runtime/src/handler/map_handler.hpp:37:11: error: unknown type name 'pthread_spinlock_t'; did you mean 'pthread_rwlock_t'?
        volatile pthread_spinlock_t &spinlock;
                 ^~~~~~~~~~~~~~~~~~
```

should i replace spinlock with rwlock? for apple systems?

Officeyutong:
> getting these now: @Officeyutong @yunwei37
> 
> ```
> bpftime-hp/runtime/src/handler/map_handler.hpp:37:11: error: unknown type name 'pthread_spinlock_t'; did you mean 'pthread_rwlock_t'?
>         volatile pthread_spinlock_t &spinlock;
>                  ^~~~~~~~~~~~~~~~~~
> ```
> 
> should i replace spinlock with rwlock? for apple systems?

You may write a spinlock. It's very simple. Don't use mutex, they are very slow

attach/CMakeLists.txt
hp77-creator: in my recent commits, I am still trying to compile it, I will rename it once it is working fine.

hp77-creator:
@Officeyutong  @yunwei37 
I am getting following errors
```

/Users/hp77/repos/bpftime-hp/runtime/src/bpftime_shm.cpp:366:10: error: use of undeclared identifier 'EPOLLIN'
                                                                        EPOLLIN,
                                                                        ^
/Users/hp77/repos/bpftime-hp/runtime/src/bpftime_shm.cpp:382:10: error: use of undeclared identifier 'EPOLLIN'
                                                                        EPOLLIN,
                                                                        ^
/Users/hp77/repos/bpftime-hp/runtime/src/bpftime_shm.cpp:400:18: error: use of undeclared identifier 'sigtimedwait'
                        if (int sig = sigtimedwait(&to_block, &sig_info, &ts);
```

It is coming from `sys/epoll` , I looked for alternatives, `kevent` is one, should we go for that for BSD specific kernels?

hp77-creator:
@Officeyutong @yunwei37 
Also getting error on
```
error: use of undeclared identifier 'CLOCK_MONOTONIC_COARSE'; did you mean '_CLOCK_MONOTONIC_RAW'?
        clock_gettime(CLOCK_MONOTONIC_COARSE, &spec);
                      ^~~~~~~~~~~~~~~~~~~~~~
```

will try to refer this: https://stackoverflow.com/questions/41509505/clock-gettime-on-macos

hp77-creator:
Also getting this:
```
error: use of undeclared identifier 'gettid'
                tid = gettid();
```

hp77-creator:
also getting this:
```
runtime/src/bpf_helper.cpp:330:14: error: use of undeclared identifier 'bpf_test_run_opts'
        LIBBPF_OPTS(bpf_test_run_opts, run_opts, .ctx_in = context,
```


do we need to include `LIBBPF_OPTS`, it is coming from `bpftime_tail_call`
what should I do with it?

Officeyutong:
> @Officeyutong @yunwei37 I am getting following errors
> 
> ```
> 
> /Users/hp77/repos/bpftime-hp/runtime/src/bpftime_shm.cpp:366:10: error: use of undeclared identifier 'EPOLLIN'
>                                                                         EPOLLIN,
>                                                                         ^
> /Users/hp77/repos/bpftime-hp/runtime/src/bpftime_shm.cpp:382:10: error: use of undeclared identifier 'EPOLLIN'
>                                                                         EPOLLIN,
>                                                                         ^
> /Users/hp77/repos/bpftime-hp/runtime/src/bpftime_shm.cpp:400:18: error: use of undeclared identifier 'sigtimedwait'
>                         if (int sig = sigtimedwait(&to_block, &sig_info, &ts);
> ```
> 
> It is coming from `sys/epoll` , I looked for alternatives, `kevent` is one, should we go for that for BSD specific kernels?

- `EPOLLIN` is a constant defined in epoll headers
- `sigtimedwait` is a POSIX signal API. Does darwin support it?

Officeyutong:
> also getting this:
> 
> ```
> runtime/src/bpf_helper.cpp:330:14: error: use of undeclared identifier 'bpf_test_run_opts'
>         LIBBPF_OPTS(bpf_test_run_opts, run_opts, .ctx_in = context,
> ```
> 
> do we need to include `LIBBPF_OPTS`, it is coming from `bpftime_tail_call` what should I do with it?

You may disable this helper function on darwin. It relies kernel eBPF to work.

hp77-creator:
> 
> * `EPOLLIN` is a constant defined in epoll headers. We didn't use epoll syscalls in bpftime, but just used constants, types, structs from epoll


I have defined the value in a separate header file for this

> * `sigtimedwait` is a POSIX signal API. Does darwin support it?

It doesn't, have created a custom implementation and added in a header file.



> 
> `gettid` returns the thread id of the currently executing thread. It's a linux specified syscall but I think there should be alternatives on darwin

Yea, I found some alternative in `pthread_threadid_np(NULL, &tid);` and using that.

> You may disable this helper function on darwin. It relies kernel eBPF to work.

I have disabled this function

I am getting a lot of error in `prog_array.cpp` now, a lot of linux/libc related functions and constants are used.

hp77-creator:
also what to do of `bpf_object.cpp` ?

hp77-creator:
I did that, Now for some reason `boost` library is not recognized
getting this error now:
```
fatal error: 'boost/interprocess/interprocess_fwd.hpp' file not found
#include <boost/interprocess/interprocess_fwd.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[3]: *** [tools/bpftimetool/CMakeFiles/bpftimetool.dir/main.cpp.o] Error 1
make[2]: *** [tools/bpftimetool/CMakeFiles/bpftimetool.dir/all] Error 2
make[1]: *** [all] Error 2
make: *** [build-wo-libbpf] Error 2
```

hp77-creator:
I have done so many changes to rectify errors now that I have lost track of my changes, will need to look at them again. 
I had faced this `boost` error before as well, setting BOOST_ROOT fixed this before. will try again.

Officeyutong:
> rectify

boost may now be recognized automically. Sometimes you may need to set BOOST_ROOT manualy

Officeyutong:
> I have done so many changes to rectify errors now that I have lost track of my changes, will need to look at them again. I had faced this `boost` error before as well, setting BOOST_ROOT fixed this before. will try again.

boost may not be recognized automically. Sometimes you may need to set BOOST_ROOT manualy

Officeyutong:
> also what to do of `bpf_object.cpp` ?

This file is mainly used for parsing ELF files. It calls API provided by libbpf to pick up eBPF programs defined in an ELF file. It may be replaced with other libraries that can parse ELF, such as libelf.

eBPF programs are stored in program sections of an ELF file. section names would be like `uprobe/XXXXX` or `tracepoint/syscall/XXXX` or other things alike. (You may use readelf to have  glance at it). Some eBPF programs might require linking with other eBPF programs (such as an eBPF program calling another non-inline function), but in my opinion we may temporarily not provide support for linking

You may refer to https://github.com/libbpf/libbpf/blob/42065ea6627ff6e1ab4c65e51042a70fbf30ff7c/src/libbpf.c#L7892 for how libbpf loads ELF files

hp77-creator:
I will read more on the above. An update that I have been able to compile bpftime successfully. But when I tried running it on my system, I am getting following error:
```
2024-06-09 02:47:53.495] [info] Injecting to 99920
[2024-06-09 02:47:53.664] [error] Failed to inject: Module not found at "/usr/lib/libz.1.dylib"
```

I checked, macOS comes with a zlib installation which is symlinked to libz.1.dylib as `libz.1.2.1.dylib`, I thought of adding one more symlink but there are various write permission issues. I will try it again.

cmake/frida.cmake
```diff
@@ -27,10 +27,10 @@ elseif(${FRIDA_OS_ARCH} MATCHES "linux-arm.*")
   # Frida only has armhf builds..
   set(FRIDA_OS_ARCH "linux-armhf")
 elseif(${FRIDA_OS_ARCH} MATCHES "darwin-arm.*")
```
viniciusd, line 29:
```suggestion
elseif(${FRIDA_OS_ARCH} MATCHES "darwin-arm64")
```

runtime/src/platform_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+    size_t len = sizeof(cpu);
+
+    if (sysctlbyname("hw.cpulocation", &cpu, &len, NULL, 0) == -1) {
+        return -1;  // Handle error
+    }
+    return cpu;
+}
+
+int sched_getaffinity(pid_t pid, size_t cpusetsize, cpu_set_t *mask) {
+    // Simulate affinity as macOS does not support setting/getting affinity
```
viniciusd, line 22: Can't we emulate it with `thread_policy_set`?

runtime/src/bpf_helper.cpp
```diff
@@ -175,7 +198,11 @@ uint64_t bpf_ktime_get_coarse_ns(uint64_t, uint64_t, uint64_t, uint64_t,
 				 uint64_t)
 {
 	struct timespec ts;
-	clock_gettime(CLOCK_MONOTONIC_COARSE, &ts);
+	#ifdef __APPLE__
+	clock_gettime(CLOCK_MONOTONIC, &ts); // or CLOCK_MONOTONIC_RAW
+	#else
+	clock_gettime(CLOCK_MONOTONIC_COARSE, &spec);
```
viniciusd, line 204:
```suggestion
	clock_gettime(CLOCK_MONOTONIC_COARSE, &ts);
```
?

runtime/src/handler/map_handler.cpp
```diff
@@ -474,6 +484,7 @@ int bpf_map_handler::map_init(managed_shared_memory &memory)
 			container_name.c_str())(memory, key_size, value_size);
 		return 0;
 	}
+	#if __linux__
```
viniciusd, line 487: `USE_BPF`?

runtime/extension/extension_helper.cpp
```diff
@@ -20,8 +20,10 @@
 #include <vector>
 
 #ifdef BPFTIME_ENABLE_IOURING_EXT
+#ifdef USE_BPF
```
viniciusd, line 23: `__linux__` instead? uring is specific to linux

runtime/agent/CMakeLists.txt
```diff
@@ -1,8 +1,13 @@
 add_library(bpftime-agent SHARED
   agent.cpp
 )
-add_dependencies(bpftime-agent FridaGum spdlog::spdlog bpftime_frida_uprobe_attach_impl bpftime_syscall_trace_attach_impl)
+if(${BPFTIME_BUILD_WITH_LIBBPF})
+  add_dependencies(bpftime-agent FridaGum spdlog::spdlog bpftime_frida_uprobe_attach_impl bpftime_syscall_trace_attach_impl)
+else()
+  add_dependencies(bpftime-agent FridaGum spdlog::spdlog bpftime_frida_uprobe_attach_impl)
```
viniciusd, line 7: Should `bpftime_frida_uprobe_attach_impl` be removed as well?

runtime/agent/CMakeLists.txt
```diff
@@ -20,5 +25,24 @@ target_link_libraries(bpftime-agent
...
+  ${FRIDA_UPROBE_ATTACH_IMPL_INCLUDE}
+)
+target_link_libraries(bpftime-agent
+  ${FRIDA_GUM_INSTALL_DIR}/libfrida-gum.a
+  runtime
+  -lpthread
+  -lm
+  -ldl
+  spdlog::spdlog
+  bpftime_frida_uprobe_attach_impl
```
viniciusd, line 46: Same as above, should `bpftime_frida_uprobe_attach_impl` be removed without bpf?

runtime/agent/CMakeLists.txt
```diff
@@ -20,5 +25,24 @@ target_link_libraries(bpftime-agent
...
 )
+
+else()
+target_include_directories(bpftime-agent
+  PRIVATE
+  ${FRIDA_GUM_INSTALL_DIR}
+  ../include
+  ../../third_party/
+  ${SPDLOG_INCLUDE}
+  ${FRIDA_UPROBE_ATTACH_IMPL_INCLUDE}
```
viniciusd, line 37: Shouldn't we keep `SYSCALL_TRACE_ATTACH_IMPL_INCLUDE`?

daemon/CMakeLists.txt
```diff
@@ -27,13 +29,21 @@ add_executable(bpftime_daemon
...
 
-add_dependencies(libbpftime_daemon
-    bpftime_daemon_ebpf_skel
-    libbpf
-    spdlog::spdlog
-    runtime
-)
+if(${BPFTIME_BUILD_WITH_LIBBPF})
+    add_dependencies(libbpftime_daemon
+        bpftime_daemon_ebpf_skel
```
viniciusd, line 34: Does `bpftime_daemon_ebpf_skel` require libbpf?

runtime/agent/CMakeLists.txt
Officeyutong: > Shouldn't we keep `SYSCALL_TRACE_ATTACH_IMPL_INCLUDE`?

It's not supported on platforms other than x86_64

viniciusd: Can't we support it for ARM?

Officeyutong: > Can't we support it for ARM?

We used a mechanism that was only designed for x86, in file https://github.com/eunomia-bpf/bpftime/blob/master/attach/text_segment_transformer/text_segment_transformer.cpp . We replace instruction `syscall` to `call *%rax` (both of them are two bytes) and setup appropriate trampoline code on page zero to implement syscall trace on x86. If we want to support syscall trace on other platforms, we need a new mechanism to implement similar thing

attach/CMakeLists.txt
hp77-creator: should I create another option for this or just define a new spec for when it is an apple system and exclude these files?

runtime/src/platform_utils.cpp
hp77-creator: i think should be done with something like that, right now it is just a stub, will explore.

runtime/src/handler/map_handler.cpp
hp77-creator: i can use that but since i have added guards for kernel maps, I preferred `__linux__` 
@Officeyutong  @yunwei37  what do you suggest?

runtime/extension/extension_helper.cpp
hp77-creator: I have added your suggestion

runtime/src/handler/map_handler.cpp
Officeyutong: > i can use that but since i have added guards for kernel maps, I preferred `__linux__` @Officeyutong @yunwei37 what do you suggest?

`USE_BPF` might be better, since kernel user shared maps relies on libbpf to work

runtime/src/platform_utils.cpp
hp77-creator: will create a new issue for this one, currently this is added as just a no operation stub.

runtime/src/handler/map_handler.cpp
hp77-creator: I have replaced __linux__ with USE_LIBBPF

runtime/agent/CMakeLists.txt
hp77-creator: discussion out of scope this PR. Resolving it, cc: @Officeyutong @viniciusd

attach/CMakeLists.txt
hp77-creator: I have used option like `if(UNIX NOT APPLE)`
let me know if you have think otherwise.

hp77-creator:
So, some guards were missing, I have added those and regression is working fine now. 
Please provide your feedback @Officeyutong @yunwei37

CMakeLists.txt
```diff
@@ -191,9 +199,19 @@ if(${BUILD_ATTACH_IMPL_EXAMPLE})
...
+endif()
 
 
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
 
-install(TARGETS bpftime-agent bpftime_text_segment_transformer bpftime-syscall-server CONFIGURATIONS Release Debug RelWithDebInfo DESTINATION ~/.bpftime)
+set(DEST_DIR "$ENV{HOME}/.bpftime")
+if(${BPFTIME_BUILD_WITH_LIBBPF})
+    install(TARGETS bpftime-agent bpftime_text_segment_transformer bpftime-syscall-server CONFIGURATIONS Release Debug RelWithDebInfo DESTINATION ${DEST_DIR})
+elseif(APPLE)
```
viniciusd, line 213: Is there any difference between the `elseif` and the `else` branch? I can't spot any

attach/CMakeLists.txt
```diff
@@ -5,5 +5,7 @@ if(CMAKE_BUILD_TYPE STREQUAL "Debug")
   endif()
 add_subdirectory(base_attach_impl)
 add_subdirectory(frida_uprobe_attach_impl)
-add_subdirectory(syscall_trace_attach_impl)
-add_subdirectory(text_segment_transformer)
+if(UNIX AND NOT APPLE)
+  add_subdirectory(syscall_trace_attach_impl)
+  add_subdirectory(text_segment_transformer)
+endif()
```
viniciusd, line 11: I understand these are gonna fail for now due to x86. But is it a compilation error or a runtime error for now?

attach/text_segment_transformer/agent-transformer.cpp
```diff
@@ -73,8 +73,7 @@ extern "C" void bpftime_agent_main(const gchar *data, gboolean *stay_resident)
 	cs_arch_register_x86();
 	bpftime::setup_syscall_tracer();
 	SPDLOG_DEBUG("Loading dynamic library..");
-	auto next_handle =
-		dlmopen(LM_ID_NEWLM, agent_so, RTLD_NOW | RTLD_LOCAL);
+	auto next_handle = dlmopen(LM_ID_NEWLM, agent_so, RTLD_NOW | RTLD_LOCAL);
```
viniciusd, line 76: Is this a formatting change only?

daemon/CMakeLists.txt
```diff
@@ -27,13 +29,21 @@ add_executable(bpftime_daemon
...
+        libbpf
+        spdlog::spdlog
+        runtime
+    )
+else()
+    add_dependencies(libbpftime_daemon
+        spdlog::spdlog
+        runtime
+    )
+endif()
```
viniciusd, line 44: Can we do...
```suggestion
f(${BPFTIME_BUILD_WITH_LIBBPF})
    add_dependencies(libbpftime_daemon
        bpftime_daemon_ebpf_skel
        libbpf
    )
endif()
add_dependencies(libbpftime_daemon
  spdlog::spdlog
  runtime
  )
```
? Not sure how `add_dependencies` work together on cmake

daemon/CMakeLists.txt
```diff
@@ -52,6 +62,23 @@ target_link_libraries(libbpftime_daemon PRIVATE
...
+)
+target_link_libraries(libbpftime_daemon PRIVATE
+    elf
+    z
+    spdlog::spdlog
+    runtime
+    PUBLIC
+    rt
+)
+endif()
```
viniciusd, line 80: Same as above... or would an empty `LIBBPF_INCLUDE_DIR` just work without branching for including directories?

runtime/src/platform_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+int my_sched_getcpu() {
+    return ::sched_getcpu();
+}
+
+#elif __APPLE__
+#include <sys/sysctl.h>
+#include <pthread.h>
+
+int my_sched_getcpu() {
```
viniciusd, line 14: Maybe... `platform_sched_getcpu()`?

runtime/src/platform_utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+int sched_setaffinity(pid_t pid, size_t cpusetsize, const cpu_set_t *mask) {
+    (void)pid;
+    (void)cpusetsize;
+    (void)mask;
+    return 0;
+}
+
+#endif
+
+int get_current_cpu() {
```
viniciusd, line 41: What's the difference between `get_current_cpu()` and `my_sched_getcpu()`? I see them being used interchangeably

runtime/src/platform_utils.cpp
viniciusd: Linux's is called `sched_getcpu()` already, right? What if we just define it for APPLE? Then it would be magically available on both, wouldn't it?

As long as we implement the same interface

tools/cli/main.cpp
```diff
@@ -219,7 +251,11 @@ int main(int argc, const char **argv)
 		return run_command(executable_path.c_str(), extra_args,
 				   so_path.c_str(), nullptr);
 	} else if (program.is_subcommand_used("start")) {
+		#if __linux__
 		auto agent_path = install_path / "libbpftime-agent.so";
+		#elif __APPLE__
+		auto agent_path = install_path / "libbpftime-agent.dylib";
+		#endif
```
viniciusd, line 258:
```suggestion
#if __linux__
  #define AGENT_LIBRARY "libbpftime-agent.so"
#elif __APPLE__
  #define AGENT_LIBRARY "libbpftime-agent.dylib"
#endif 
auto agent_path = install_path / AGENT_LIBRARY;
```
The `AGENT_LIBRARY` definition would probably go higher up in the file and be reused below as well.

What do you think?

CMakeLists.txt
```diff
@@ -191,9 +199,19 @@ if(${BUILD_ATTACH_IMPL_EXAMPLE})
...
 add_subdirectory(benchmark)
+endif()
 
 
 set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>DLL")
 
-install(TARGETS bpftime-agent bpftime_text_segment_transformer bpftime-syscall-server CONFIGURATIONS Release Debug RelWithDebInfo DESTINATION ~/.bpftime)
+set(DEST_DIR "$ENV{HOME}/.bpftime")
+if(${BPFTIME_BUILD_WITH_LIBBPF})
+    install(TARGETS bpftime-agent bpftime_text_segment_transformer bpftime-syscall-server CONFIGURATIONS Release Debug RelWithDebInfo DESTINATION ${DEST_DIR})
```
Officeyutong, line 212: Syscall server shouldn't be excluded if libbpf was not enabled. bpftime relies on syscall server to work, by injecting syscall server to a userspace eBPF program. If some parts of syscall server won't work on macos, try adapting or disabling these parts

attach/CMakeLists.txt
Officeyutong: A runtime error
- text segment transformer emits x86 assembly, these instructions are obviously illegal on aarch64

daemon/CMakeLists.txt
```diff
@@ -1,10 +1,12 @@
-# Create a target that builds the ebpf program
-add_ebpf_program_target(bpftime_daemon_ebpf_target ${CMAKE_CURRENT_SOURCE_DIR}/kernel/bpf_tracer.bpf.c ${CMAKE_CURRENT_BINARY_DIR}/bpf_tracer.bpf.o)
+if(${BPFTIME_BUILD_WITH_LIBBPF})
```
Officeyutong, line 1: bpftime_daemon won't work on platforms other than Linux. The whole daemon folder could be disabled. It relies on kernel eBPF to work. 
- bpftime_daemon acts similarly like syscall server, but it works by monitoring syscalls like `bpf` in kernel by kernel eBPF and sends such calling records to user space and stores it to shared memory. While syscall server hook such calls in userspace

runtime/CMakeLists.txt
```diff
@@ -122,15 +123,21 @@ target_include_directories(${PROJECT_NAME}
   ${CMAKE_CURRENT_BINARY_DIR}
   ${CMAKE_CURRENT_SOURCE_DIR}/../third_party
   ${SPDLOG_INCLUDE}
+  ${Boost_INCLUDE_DIRS}
 )
 
+message(INFO " Found the following boost link dirs : ${Boost_LIBRARIES}")
```
Officeyutong, line 129: `link` or `lib`?

runtime/include/spinlock_wrapper.hpp
```diff
@@ -0,0 +1,75 @@
...
+// Include the custom spinlock implementation
+#include "spinlock.hpp"
+
+// Define the custom spinlock type and functions only if the standard library is not available
+#if !defined(_POSIX_SPIN_LOCKS) || _POSIX_SPIN_LOCKS <= 0
+typedef Spinlock pthread_spinlock_t; // Define custom spinlock type
+
+// Non-volatile version
+inline int pthread_spin_init(pthread_spinlock_t* lock, int pshared) {
+    (void)pshared; // suppress unused parameter warning
```
Officeyutong, line 17: What about using `[[maybe_unused]]`? 

See https://en.cppreference.com/w/cpp/language/attributes/maybe_unused

runtime/src/bpf_helper.cpp
```diff
@@ -727,7 +758,7 @@ const bpftime_helper_group shm_maps_group = { {
 } };
 
 extern const bpftime_helper_group extesion_group;
-
+#if __linux
```
Officeyutong, line 761: `__linux` or `__linux__`? If both of them are available, only choose one option

Officeyutong (changes requested):

attach/CMakeLists.txt
Officeyutong: > I understand these are gonna fail for now due to x86. But is it a compilation error or a runtime error for now?

A runtime error

text segment transformer emits x86 assembly now, and the corresponding trampoline code was designed for x86

CMakeLists.txt
hp77-creator: i had it written for a different rule for APPLE, but by the end it was same for else and elseif, i will combine them. Thanks for pointing out.

attach/text_segment_transformer/agent-transformer.cpp
hp77-creator: i had added APPLE specific code before, but since I had removed the `add_subdirectory` for this folder, I removed the guards, what remain was the formatting change.

CMakeLists.txt
```diff
@@ -155,7 +161,9 @@ endif()
 bpftime_add_libs_component_command(${FRIDA_GUM_INSTALL_DIR}/libfrida-gum.a)
 bpftime_add_libs_component_command(${CMAKE_CURRENT_BUILD_DIR}/libbpf/libbpf/libbpf.a)
```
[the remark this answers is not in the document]
hp77-creator: noted, will do this change.

CMakeLists.txt
hp77-creator: okay, will have to add those changes then.

attach/CMakeLists.txt
hp77-creator: it fails during `cmake --build build` as well because headers are not defined and there are a lot of headers which will need to be changed.

daemon/CMakeLists.txt
hp77-creator: noted, will do that change.

daemon/CMakeLists.txt
hp77-creator: didn't understand this change.

runtime/CMakeLists.txt
hp77-creator: they are added in linking step so 'link' also it was for debug, I will remove it,

runtime/include/spinlock_wrapper.hpp
hp77-creator: okay, will incorporate the above change

runtime/src/bpf_helper.cpp
hp77-creator: should be `__linux__` it's a typo. Thanks for pointing out.

runtime/src/platform_utils.cpp
hp77-creator: I tried that way but was facing warning's from gcc on infinite recursion, even though it was two different scoped functions, you can see in previous version, it was 'sched_getcpu()' with consistent interface, but i replaced it altogether to remove those warnings. @Officeyutong  @yunwei37 what do you suggest?

tools/cli/main.cpp
hp77-creator: yea, agreed, should have it in global space and replace where it is being used by the defined variable.

runtime/src/platform_utils.cpp
hp77-creator: you are right, there should be only one, i will update the header. Thanks.

runtime/src/platform_utils.cpp
hp77-creator: using `get_currentcpu()` in all files now.

daemon/CMakeLists.txt
hp77-creator: have removed flag from this `CMakeLists.txt` file.

hp77-creator:
Also @Officeyutong @yunwei37 
currently i have added all headers needed for `bpf` and `epoll` in `bpftime_poll.h`, I was not sure what header name to keep or how to differentiate so I have added all `structs` and everything into one, if it works, we can merge it like this and then as @viniciusd includes support for frida and syscalls we can separate out what is needed and what is not.

hp77-creator:
@Officeyutong @yunwei37 should we also create a CI for macOS?

Officeyutong:
> @Officeyutong @yunwei37 should we also create a CI for macOS?
> 
> 

Good idea! We may migrate frida attach manager tests CI to macOS first

runtime/include/spinlock.hpp
```diff
@@ -0,0 +1,59 @@
...
+        }
+        return *this;
+    }
+
+    void spin_lock() {
+        static const timespec ns = {0, 1};
+        for(int i = 0; lock.load(std::memory_order_relaxed) || lock.exchange(1, std::memory_order_acquire); ++i) {
+            if(i==8){
+                i=0;
+                nanosleep(&ns, NULL);
```
Officeyutong, line 40: Nanosleep is a syscall. It would be better if there is a pure user space spinlock implementation. Syscall are very slow and we should avoid them in user space eBPF.

If there is any similar instruction like pause on x86, replace the loop body with it using inline assembly. If not, just leave an empty loop body

runtime/include/spinlock.hpp
```diff
@@ -0,0 +1,59 @@
...
+            if (other.lock.exchange(1, std::memory_order_acquire)) {
+                lock.store(0, std::memory_order_release);
+            }
+        }
+        return *this;
+    }
+
+    void spin_lock() {
+        static const timespec ns = {0, 1};
+        for(int i = 0; lock.load(std::memory_order_relaxed) || lock.exchange(1, std::memory_order_acquire); ++i) {
```
Officeyutong, line 37: Variable `i` will overflow. Overflow of signed integers are undefined behaviors. Undefined behaviors may lead to incorrect compiler optimizations

https://stackoverflow.com/questions/16188263/is-signed-integer-overflow-still-undefined-behavior-in-c

Officeyutong (changes requested):

attach/CMakeLists.txt
viniciusd: I asked it because, the way I see it, the syscall trace component is about supporting ARM overall, not an Apple-specific feature. We should support it eventually and that's orthogonal to this PR, which is about making libbpf optional.

Mac support will benefit from ARM support, but:
1. Linux ARM support also will
2. By removing libbpf, maybe we can get it working on Mac x86? Aka non-M chips and/or via rosetta

runtime/include/spinlock.hpp
viniciusd: For future reference, keep in mind that compilers re-order instructions at will, that's why it is common to have a memory barrier in lock implementations.

I see you are using the relaxed memory order, but I am not confident it will do the trick. Keep it in mind when/if implementing the in-line asm version as well. I am particularly not sure if/how compilers also move in-line ASM around.

https://en.cppreference.com/w/cpp/atomic/memory_order

attach/CMakeLists.txt
hp77-creator: Although PR started as just adding an option for disabling support for `libbpf`, I have tried to compile it on `aarch64` systems. We can create an issue which will remove this particular check by making it compatible for all systems but given our ultimate goal of making it run on macOS, I took one step in that direction, although for macOS systems with m* chip series in this case. Above check will disable `syscall_trace_attach_impl` and `text_segment_transformer` for APPLE systems.
As for rosetta systems, it is another setup on system and `brew` messes up architectures for packages, so don't wanna go through that. Maybe you can try it and let us know.

runtime/include/spinlock.hpp
hp77-creator: will it overflow though? there is a check for `i==8`?

runtime/include/spinlock.hpp
hp77-creator: Noted @Officeyutong 


> For future reference, keep in mind that compilers re-order instructions at will, that's why it is common to have a memory barrier in lock implementations.

Thanks for telling me about this. 

> I see you are using the relaxed memory order, but I am not confident it will do the trick. Keep it in mind when/if implementing the in-line asm version as well. I am particularly not sure if/how compilers also move in-line ASM around.

I have seen it in many blogs being used, so I am guessing it should work, but yea I will look into it, as for in-line asm, will check, what can I replace it with, Thanks for your input.

> https://en.cppreference.com/w/cpp/atomic/memory_order
Thanks for the reference

runtime/include/spinlock.hpp
Officeyutong: My fault. Sorry

Officeyutong:
> > > @Officeyutong @yunwei37 should we also create a CI for macOS?
> > 
> > 
> > Good idea! We may migrate frida attach manager tests CI to macOS first
> 
> That's related to something I wanted to bring up in #304, but I was trying to make sense of the current implementation and reviewing the paper before asking:
> 
> `bpftime_frida_uprobe_attach_tests` doesn't break at the moment. All tests pass even though I am on ARM:
> 
> ```
> $ ./bpftime/build/attach/frida_uprobe_attach_impl/bpftime_frida_uprobe_attach_tests
> Randomness seeded to: 3365707513
> ===============================================================================
> All tests passed (1025 assertions in 8 test cases)
> ```
> 
> They were expected to fail/break, right?

It was not expected to fail. But we have never tested it on macos

hp77-creator:
> > @Officeyutong @yunwei37 should we also create a CI for macOS?
> > 
> > 
> 
> Good idea! We may migrate frida attach manager tests CI to macOS first

Alright,
we can take it up in another PR then
Will fix this PR with requested changes and we can merge it.

runtime/include/spinlock.hpp
hp77-creator: This https://github.com/gstrauss/plasma/blob/3d5b64539f6aba1218e303be7e33f47c6ad74c22/plasma_spin.h#L94 
has used `yield`

runtime/include/bpftime_prog.hpp
```diff
@@ -13,7 +13,11 @@
 namespace bpftime
 {
 
+#ifdef __APPLE__
+extern __thread std::optional<uint64_t> current_thread_bpf_cookie;
```
Officeyutong, line 17: Why using __thread instead of thread_local?

runtime/include/spinlock.hpp
```diff
@@ -0,0 +1,63 @@
...
+            if (other.lock.exchange(1, std::memory_order_acquire)) {
+                lock.store(0, std::memory_order_release);
+            }
+        }
+        return *this;
+    }
+
+    void spin_lock() {
+        for(int i = 0; lock.load(std::memory_order_relaxed) || lock.exchange(1, std::memory_order_acquire); ++i) {
+            if(i==8){
```
Officeyutong, line 37: I think there is no need for checking every 8 bumps, it can just be a while loop and calling yield everytime

runtime/src/bpf_map/userspace/prog_array.cpp
```diff
@@ -3,22 +3,30 @@
  * Copyright (c) 2022, eunomia-bpf org
  * All rights reserved.
  */
+
+#if __linux__
 #include "bpf/bpf.h"
 #include "linux/bpf.h"
+#include <bpf/libbpf.h>
+#include <gnu/lib-names.h>
```
Officeyutong, line 11: Which library is this header in? glibc or other lib?

runtime/include/spinlock.hpp
```diff
@@ -0,0 +1,63 @@
...
+                i=0;
+                #if __APPLE__ 
+                __asm__ __volatile__ ("yield");
+                #else 
+                __asm__ __volatile__ ("pause")
```
Officeyutong, line 42: Why calling pause when __APPLE__ is not defined?

runtime/src/platform_utils.cpp
```diff
@@ -0,0 +1,40 @@
...
+#elif __APPLE__
+#include <sys/sysctl.h>
+#include <pthread.h>
+
+int my_sched_getcpu() {
+    int cpu = -1;
+    size_t len = sizeof(cpu);
+
+    if (sysctlbyname("hw.cpulocation", &cpu, &len, NULL, 0) == -1) {
+        return -1;  // Handle error
```
Officeyutong, line 19: Call SPDLOG_ERROR to log errors

tools/cli/main.cpp
```diff
@@ -19,6 +19,44 @@
 #include <tuple>
 #include <sys/wait.h>
 
+#ifdef __APPLE__
+#include <crt_externs.h>
+#include <cstdlib>
+#include <cerrno>
+#include <cstring>
+#include <unistd.h>
+#define environ (*_NSGetEnviron())
```
Officeyutong, line 28: Avoid using #define to define inline functions in C++, use inline function instead. Believe optimization by compiler

runtime/src/bpf_map/userspace/prog_array.cpp
hp77-creator: I guess, I added this, because of some references were not working on my system, i guess these are not needed since those errors might have crept up because of other libraries. I will check it.

runtime/include/spinlock.hpp
hp77-creator: forgot that this is only for APPLE, will remove this guard, thanks.

runtime/src/platform_utils.cpp
hp77-creator: okay, will do this change

tools/cli/main.cpp
```diff
@@ -19,6 +19,44 @@
...
+#define AGENT_LIBRARY "libbpftime-agent.dylib"
+#define SYSCALL_SERVER_LIBRARY "libbpftime-syscall-server.dylib"
```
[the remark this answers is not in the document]
hp77-creator: okay, will replace.

runtime/include/bpftime_prog.hpp
hp77-creator: thread_local was not present in APPLE, i was giving errors when trying to compile it.

runtime/include/spinlock.hpp
hp77-creator: alright.

runtime/src/bpf_map/userspace/prog_array.cpp
hp77-creator: CI failed after I remove 'gnu/lib-names', https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.7-4.6/+/jb-dev/sysroot/usr/include/gnu/lib-names.h
it is part of glibc, why is it failing now? but earlier CI used to succeed even without gnu/lib-names? should I include it now?

runtime/include/bpftime_prog.hpp
Officeyutong: > thread_local was not present in APPLE, i was giving errors when trying to compile it.

`thread_local` is a keyword introduced in C++11, you may need to check your compiler

https://en.cppreference.com/w/cpp/keyword/thread_local

runtime/src/bpf_map/userspace/prog_array.cpp
Officeyutong: > CI failed after I remove 'gnu/lib-names', https://android.googlesource.com/platform/prebuilts/gcc/linux-x86/host/x86_64-linux-glibc2.7-4.6/+/jb-dev/sysroot/usr/include/gnu/lib-names.h it is part of glibc, why is it failing now? but earlier CI used to succeed even without gnu/lib-names? should I include it now?

Okay, then let's keep that include

tools/cli/main.cpp
hp77-creator: done.

runtime/src/platform_utils.cpp
hp77-creator: have added a `SPDLOG_ERROR` and am also returning -1 to be handled in other functions wherever it is called.

runtime/src/platform_utils.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    if (sysctlbyname("hw.cpulocation", &cpu, &len, NULL, 0) == -1) {
+        SPDLOG_ERROR("Couldn't get cpu location for the system");
+        return -1;  
+    }
+    return cpu;
+}
+
+int sched_getaffinity(pid_t pid, size_t cpusetsize, cpu_set_t *mask) {
+    (void)pid;
```
Officeyutong, line 27: All unused parameters should be marked with [[maybe_unused]]

tools/bpftimetool/main.cpp
```diff
@@ -83,8 +83,11 @@ int bpftime_run_ebpf_program(int id,
 			auto new_prog = bpftime_prog(prog.insns.data(),
 							 prog.insns.size(),
 							 prog.name.c_str());
+			// kernel function can only be used in linux env
+			#if __linux__
 					bpftime::bpftime_helper_group::get_kernel_utils_helper_group()
```
Officeyutong, line 88: `kernel_utils_helper` here means kernel eBPF helpers implemented in userspace by bpftime. Most of them are platform independent and able to run macos. Don't exclude the whole helper group on linux platforms

Officeyutong (changes requested):

benchmark/test_embed.c
```diff
@@ -7,8 +7,10 @@
 #include <stdio.h>
```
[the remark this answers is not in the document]
hp77-creator: these changes were done before i discovered that benchmark is only for libbpf, i can remove these guards.

runtime/src/platform_utils.cpp
hp77-creator: sorry, will take care of these.

tools/bpftimetool/main.cpp
hp77-creator: I see, will go through these changes then

Makefile
```diff
@@ -54,6 +54,10 @@ build-iouring: ## build the package with iouring extension
 	cmake -Bbuild -DBPFTIME_ENABLE_IOURING_EXT=1 -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo
 	cmake --build build --config RelWithDebInfo  -j$(JOBS)
 
+build-wo-libbpf: ## build the package with iouring extension
+	cmake -Bbuild -DBPFTIME_ENABLE_IOURING_EXT=1 -DCMAKE_BUILD_TYPE:STRING=RelWithDebInfo -DBPFTIME_BUILD_WITH_LIBBPF=OFF -DBPFTIME_ENABLE_IOURING_EXT=OFF -DBPFTIME_BUILD_KERNEL_BPF=OFF
```
Officeyutong, line 58: Option `BPFTIME_ENABLE_IOURING_EXT` occurred two times

runtime/include/platform_utils.hpp
```diff
@@ -0,0 +1,31 @@
+#ifndef PLATFORM_UTIL_H
+#define PLATFORM_UTIL_H
+#include <cstdlib>
+#include <functional>
+
+#if __linux__
+    #include <sched.h>
```
Officeyutong, line 7: Strange indent

runtime/extension/extension_helper.cpp
```diff
@@ -46,8 +48,13 @@ uint64_t bpftime_path_join(const char *filename1, const char *filename2,
 
 namespace bpftime
 {
-#ifdef BPFTIME_ENABLE_IOURING_EXT
+/*
+io_uring are only available in linux atm 
+so adding linux guards to the following code
+*/
 
+#ifdef BPFTIME_ENABLE_IOURING_EXT
```
Officeyutong, line 56: Nested `#ifdef - #if` could be replaced with `#if defined(XXX) && YYY` to make it clearer

runtime/src/bpf_helper.cpp
```diff
@@ -3,14 +3,22 @@
  * Copyright (c) 2022, eunomia-bpf org
  * All rights reserved.
  */
+#if __APPLE__
+#include <_types/_uint64_t.h>
```
Officeyutong, line 7: What's the purpose of this header?

Officeyutong (changes requested):

tools/cli/main.cpp
```diff
@@ -19,6 +19,46 @@
...
+        buf[len] = '/';
+        strcpy(buf + len + 1, file);
+        execve(buf, argv, envp);
+        if (errno != ENOENT)
+            return -1;
+    }
+    errno = ENOENT;
+    return -1;
+}
+#else
```
viniciusd, line 55: Can we do `#elif __linux__` so it fails on other platforms that haven't been implemented?

Makefile
hp77-creator: both are not required, will remove them.

runtime/src/bpf_helper.cpp
hp77-creator: had added it before to use `uint64` but later on made my custom implementation. removing it.

runtime/CMakeLists.txt
```diff
@@ -188,15 +221,28 @@ target_include_directories(${PROJECT_NAME} PUBLIC
...
+)
+endif()
 
 message(DEBUG "Successfully added all dependencies and linked against them.")
 
 set(BPFTIME_RUNTIME_INCLUDE ${CMAKE_CURRENT_SOURCE_DIR}/include ${CMAKE_CURRENT_SOURCE_DIR}/src)
 
+
+if(UNIX AND NOT APPLE)
 add_subdirectory(object)
```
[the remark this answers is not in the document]
hp77-creator: i don't understand? do u mean exclude for all platforms or just for APPLE systems?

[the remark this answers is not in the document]
Officeyutong: > i don't understand? do u mean exclude for all platforms or just for APPLE systems?

I remember that some parts of bpftime depends on the object folder, right? If we exclude that folder, will these parts still work?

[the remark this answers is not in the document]
hp77-creator: I was getting a lot of errors from object, so I removed it and i think u only suggested, i will check which ones do we need to remove or what changes we will have to make.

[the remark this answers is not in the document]
hp77-creator: I have added definitions for structure and functions which were failing from libbpf header file, they are passing now without the guard.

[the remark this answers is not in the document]
Officeyutong: Okay. I've looked at code of bpftime and found that only some unit tests (in target `bpftime_runtime_tests`) depends on bpftime_object. Remember to write a custom implementation of bpftime_object based on libelf in the future if we want to make unit tests work on macos

Officeyutong (approved):

Officeyutong:
@viniciusd Remeber to approve this PR if there is no other issues. I'll merge this once you approve it

Officeyutong:
Remeber to resolve conflicts @hp77-creator

hp77-creator:
Alright resolving.

Officeyutong:
Also, options added should be documented in https://github.com/eunomia-bpf/eunomia.dev/blob/main/docs/bpftime/documents/build-and-test.md, remember to update it

@hp77-creator

hp77-creator:
Will do

PR #321, opened 2024-08-10
fix: improve error message and handling in syscall-server

yunwei37:
## Description

- Improve error message when attach to unrelated process
- Improve syscall server error handle

Now, if one uprobe attach failed, it will continue trying to attach other uprobes.

### DIFF
--- a/attach/frida_uprobe_attach_impl/src/frida_attach_utils.cpp
+++ b/attach/frida_uprobe_attach_impl/src/frida_attach_utils.cpp
@@ -11,23 +11,23 @@ static std::string get_executable_path()
 {
 	char exec_path[PATH_MAX] = { 0 };
 
-	#if __linux__
-		ssize_t len =
-			readlink("/proc/self/exe", exec_path, sizeof(exec_path) - 1);
-		if (len != -1) {
-			exec_path[len] = '\0'; // Null-terminate the string
-			SPDLOG_INFO("Executable path: {}", exec_path);
-		} else {
-			SPDLOG_ERROR("Error retrieving executable path: {}", errno);
-		}
-	#elif __APPLE__
-		pid_t pid = getpid();
-		if (proc_pidpath(pid, exec_path, sizeof(exec_path)) > 0) {
-			SPDLOG_INFO("Executable path: {}", exec_path);
-		} else {
-			SPDLOG_ERROR("Error retrieving executable path: {}", errno);
-		}
-	#endif 
+#if __linux__
+	ssize_t len =
+		readlink("/proc/self/exe", exec_path, sizeof(exec_path) - 1);
+	if (len != -1) {
+		exec_path[len] = '\0'; // Null-terminate the string
+		SPDLOG_INFO("Executable path: {}", exec_path);
+	} else {
+		SPDLOG_ERROR("Error retrieving executable path: {}", errno);
+	}
+#elif __APPLE__
+	pid_t pid = getpid();
+	if (proc_pidpath(pid, exec_path, sizeof(exec_path)) > 0) {
+		SPDLOG_INFO("Executable path: {}", exec_path);
+	} else {
+		SPDLOG_ERROR("Error retrieving executable path: {}", errno);
+	}
+#endif
 	return exec_path;
 }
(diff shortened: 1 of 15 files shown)


attach/frida_uprobe_attach_impl/src/frida_attach_utils.cpp
```diff
@@ -47,8 +47,10 @@ resolve_function_addr_by_module_offset(const std::string_view &module_name,
 			get_module_base_addr(std::string(module_name).c_str());
 	}
 	if (!module_base_addr) {
-		SPDLOG_ERROR("Failed to find module base address for {}",
-			     module_name);
+		// It's not a bug, it might be attach to a unrelated process
```
Officeyutong, line 50: Warning will be better? Attaching to unrelated processes is not expected

runtime/syscall-server/syscall_server_main.cpp
```diff
@@ -21,49 +21,83 @@ using namespace bpftime;
 // global context for bpf syscall server
 static syscall_context context;
 
+template <typename F, typename... Args>
+auto handle_exceptions(F &&f, Args &&...args) noexcept
+	-> decltype(f(std::forward<Args>(args)...))
+{
+	try {
```
Officeyutong, line 28: Will try-catch on such critical path make hooking more slower?

attach/frida_uprobe_attach_impl/src/frida_attach_utils.cpp
yunwei37: Since we are using LD_PRELOAD, there might be usecases where the user sets it as env variables to trace all commands in a bash, or the subprocesses of the program, e.g. `perf`



It's not an error or warning in these cases.

runtime/syscall-server/syscall_server_main.cpp
yunwei37: Thr syscall server is not as performance critical as the runtime. So I think it's OK

benchmark/run_benchmark.py
```diff
@@ -90,7 +90,7 @@ async def run_userspace_uprobe_test():
     await server_start_cb.wait()
     print("server started!")
     result = None
-    for i in range(100):
+    for i in range(10):
```
hp77-creator, line 93: Wouldn't reducing number of iterations reduce the reliability of our results?

yunwei37: No. it will not bring too much difference.

For example, the uprobe time in my machine is about 1000ns (kernel 10000ns)

When we have a true value of 1000, with each measurement having an error (standard deviation) of approximately 100, the measurement results can be assumed to follow a normal distribution \( X \sim N(1000, 100^2) \).

### Standard Error of the Mean
The standard error (SE) of the sample mean decreases as the sample size increases. The formula for SE is:

\[
\text{SE} = \frac{\sigma}{\sqrt{n}}
\]

where:
- \( \sigma = 100 \) (the standard deviation of measurement errors)
- \( n \) is the sample size.

- For a sample size of 10, the SE is approximately 31.62.
- For a sample size of 100, the SE reduces to 10.

### Error Comparison

- **With 10 samples**: The standard error of the mean is 31.62. This means the sample mean is expected to be within approximately ±31.62 of the true value (1000).
- **With 100 samples**: The standard error of the mean is 10, meaning the sample mean is expected to be within ±10 of the true value.

### Conclusion
While increasing the sample size from 10 to 100 does reduce the standard error, the reduction is not drastically large. The standard error decreases from 31.62 to 10, which improves the precision of the sample mean but does not imply a massive difference in accuracy. Thus, the error in estimating the true value with 10 measurements versus 100 measurements is somewhat significant but not overwhelmingly so. In practical terms, even with 10 measurements, the sample mean is expected to be reasonably close to the true value.

Officeyutong (approved):

yunwei37:
Thanks for your help!

benchmark/run_benchmark.py
hp77-creator: Thank you so much for this detailed explanation. Got it.

PR #323, opened 2024-08-10
llvm-jit: patch lddw helpers at compile time

Officeyutong:
This PR updated llvm-jit so it will execute lddw helpers (https://docs.kernel.org/bpf/standardization/instruction-set.html#id33) at runtime and hardcodes generated values into LLVM IR, thus reducing runtime oveahead

Also set default implementation for map_by_fd and map_by_idx: returning its input value

- One exception is that map_val won't be patched for AOT compilations, since AOT-ed programs could be migrated among processed, and map_val (a memory address) may be invalided among processes

### DIFF
--- a/vm/llvm-jit/src/compat/compat_llvm.cpp
+++ b/vm/llvm-jit/src/compat/compat_llvm.cpp
@@ -1,10 +1,39 @@
 #include "spdlog/spdlog.h"
 #include <bpftime_vm_compat.hpp>
 #include <cerrno>
+#include <cstdint>
 #include <memory>
 #include <ebpf_inst.h>
 #include "compat_llvm.hpp"
 #include "../llvm/llvm_jit_context.hpp"
+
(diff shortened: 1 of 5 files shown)


vm/llvm-jit/src/compat/compat_llvm.cpp
```diff
@@ -1,10 +1,39 @@
...
 #include <cerrno>
+#include <cstdint>
 #include <memory>
 #include <ebpf_inst.h>
 #include "compat_llvm.hpp"
 #include "../llvm/llvm_jit_context.hpp"
+
+static uint64_t default_map_by_fd(uint32_t)
+{
+	SPDLOG_WARN("map_by_fd not defined, but called");
```
yunwei37, line 12: I don't think we need to warn here, we can make the input val the same as output.

yunwei37 (approved):

PR #328, opened 2024-08-13
runtime: add uprobe_multi support

### LINKED ISSUE #214: [idea]: Support uprobe_multi for bpftime

Officeyutong:
uprobe_multi is a special type of uprobe, which allows hooking multiple userspace functions in a single attach. This PR adds uprobe_multi support for uprobe_attach_impl, the corresponding example, and update CI

Closes #214 

See https://lore.kernel.org/bpf/20230424160447.2005755-1-jolsa@kernel.org/ for details

## Our design

In the kernel bpf, the attaching of uprobe_multi were implemented through creating a bpf link, and filling all information (such as pid, function offset, attach cookies) in the link opts. You may find it in bpf_program__attach_uprobe_multi

But in bpftime, all attaches are required to have an attach target (a.k.a perf event), so for uprobe_multi that doesn't give us a perf event, we have a slightly different implementation to the kernel: When creating a bpf link, we doesn't only record configurations from bpf_link_create_opts into the handler, but also do we create some perf events for the uprobe_multi links, and record them in the handler. In this way, we tear down a uprobe_multi into several simple uprobe attach targets, and with a total bpf link. This will not affect the hooking performance.

## Main changes
- Update bpf_link_handler to make it support multiple attach targets
- Adapt for multiple attach targets in a single bpf_link


runtime/src/attach/bpf_attach_ctx.cpp
```diff
@@ -242,81 +240,122 @@ int bpf_attach_ctx::instantiate_prog_handler_at(int id,
...
+		// been registered
+		if (auto itr = attach_impls.find(
+			    (int)bpf_event_type::BPF_TYPE_UPROBE);
+		    itr != attach_impls.end()) {
+			std::tie(attach_impl, priv_creator) = itr->second;
+			attach_type = (int)bpf_event_type::BPF_TYPE_UPROBE;
+
+		} else {
+			SPDLOG_ERROR(
+				"Trying to instantiate uprobe multi (entry hook), but uretprobe is not registered");
```
maxgio92, line 284:
```suggestion
				"Trying to instantiate uprobe multi (entry hook), but uprobe is not registered");
```

Shouldn't the message be that way?

attach/frida_uprobe_attach_impl/src/frida_attach_private_data.cpp
```diff
@@ -28,8 +28,13 @@ int frida_attach_private_data::initialize_from_string(const std::string_view &sv
...
-		this->module_name = module_part;
 	}
 
 	return 0;
 }
+
+std::string frida_attach_private_data::to_string()
+{
+	return std::string("<Frida attach private data addr=") +
+	       std::to_string(addr);
```
maxgio92, line 39:
```suggestion
	return std::string("<Frida attach private data addr=") +
	       std::to_string(reinterpret_cast<uintptr_t>(addr)) + ">";
```

runtime/src/attach/bpf_attach_ctx.cpp
```diff
@@ -242,81 +240,122 @@ int bpf_attach_ctx::instantiate_prog_handler_at(int id,
...
+	int id, const bpf_link_handler &handler, const handler_manager *manager)
+{
+	SPDLOG_DEBUG(
+		"Instantiating bpf link handler of type perf event at id {}, prog {}, target {}",
+		id, handler.prog_id, handler.target_id);
+	auto prog = instantiated_progs.at(handler.prog_id).get();
+	// For perf event link, there should be an instantiated target
+	auto &[priv_data, attach_type] =
+		instantiated_perf_events[handler.target_id];
+	SPDLOG_DEBUG("Attach private data is {}, attach type is ",
```
maxgio92, line 326:
```suggestion
	SPDLOG_DEBUG("Attach private data is {}, attach type is {}",
```

runtime/src/attach/bpf_attach_ctx.cpp
```diff
@@ -242,81 +240,122 @@ int bpf_attach_ctx::instantiate_prog_handler_at(int id,
...
+
+		int attach_id = attach_impl->create_attach_with_ebpf_callback(
 			[=](void *mem, size_t mem_size, uint64_t *ret) -> int {
 				current_thread_bpf_cookie = cookie;
 				int err = prog->bpftime_prog_exec(
 					(void *)mem, mem_size, ret);
 				return err;
 			},
 			*priv_data, attach_type);
+		native_ids.emplace_back(attach_id, attach_impl);
```
maxgio92, line 311: Don't we need to check if the attach failed, on `attach_id`?

yunwei37:
Update: this is now addressed on the current head. The branch has been rebased, the review threads are resolved, the `attach_id` failure path now rolls back partial uprobe_multi attaches, and follow-up fixes preserve legacy `target_fd` JSON import/export and avoid deleting `BPF_TRACE_UPROBE_MULTI` links when an unrelated perf-event fd is closed. Local targeted runtime, daemon, uprobe_multi, and nginx attach checks passed; GitHub checks are rerunning on the latest push.

runtime/src/attach/bpf_attach_ctx.cpp
SHTUSIST: Handled on the current head: `attach_id` is checked before appending to `native_ids`; on failure the code logs the error, calls `rollback_native_ids()`, and returns `{}` before `native_ids.emplace_back(...)`.

yunwei37 (commented):
Pushed `1e42178` on top of the contributor branch’s latest `6343a86` without replacing its new hardening commits.

This closes the two concrete CI regressions: the daemon test now reads the cookie from `perf_event_link_data`, and custom attach restoration no longer calls the optional/throwing `private_data::to_string()` just for debug formatting. The newer target bounds/type checks and `instantiated_perf_events.find()` guard remain intact.

Validation on the exact head:
- fresh Debug build of `bpftime_daemon_tests`
- daemon suite: 2 cases / 4 assertions passed
- complete `attach_impl_example_nginx` build, including nginx 1.22.1 and the plugin adaptor
- `git diff --check`

A dedicated read-only review and an independent Kimi K3 read-only review both passed for `1e42178`. The four historical review threads are resolved; no unresolved review thread remains. GitHub CI has restarted for the pushed head. No merge was performed.

yunwei37 (approved):
Approved on exact head `1e42178` after the rebased composition was rebuilt and re-reviewed. The contributor branch’s newer hardening commits remain intact, the daemon cookie regression and custom attach restoration failure are fixed, all historical threads are resolved, and both dedicated and independent Kimi K3 read-only reviews passed. CI is still running; no merge was performed.

yunwei37 (approved):
Approved on exact head `e1c34844d4a4a8741b0611ebc624879654be1f99`. The new commit restores real legacy `target_fd` JSON compatibility, keeps current `target_id` import/export, defaults old records to `BPF_PERF_EVENT`, and prevents an unrelated perf-fd close from deleting a `BPF_TRACE_UPROBE_MULTI` link with the same numeric id.

Focused validation passed: legacy compatibility (11 assertions), uprobe_multi lifetime (9), and uprobe_multi JSON round-trip (15). A dedicated exact-head review and an independent cross-provider read-only exact-head review both passed. No review thread is unresolved. CI matrices are still running; no merge was performed.

yunwei37:
Thanks again for the uprobe_multi work — this remains in good shape with the approvals and green exact-head checks preserved.

Since the last rebase, several runtime fixes landed on `master` (#650, #656, #657, #658, #659, #662), and the branch now conflicts with current `master` in exactly two files:

- `runtime/src/attach/bpf_attach_ctx.cpp`
- `runtime/syscall-server/syscall_context.cpp`

Could you merge current `master` into the branch (or rebase) and push? The landed changes in those areas are the Frida map_files/VMA split and x86 `pt_regs` fixes plus the daemon capture rejection, so the conflicts should stay localized to your attach/syscall-server context changes. Once the merge is clean, exact-head CI will re-run and this should be ready for a maintainer merge decision.

SHTUSIST:
> **Note:** this is an AI-written reply prepared on behalf of @SHTUSIST (huyp), posted for reference. Every claim below was read from the tree at the SHAs named, but please verify before acting on it.

Checked the merge against master `b87f9682` (2026-09-04) with PR head `e1c34844` and merge base `1dd02f76`. Two conflicted files, exactly as you said.

One correction on where they come from, because it changes what to read. None of #650, #656, #657, #658, #659 or #662 touches either conflicting file. Since the merge base:

- `runtime/src/attach/bpf_attach_ctx.cpp` — exactly one commit, `aa5be1f8` (#644, "fix: correct Frida probe and tail-call runtime behavior", 2026-08-14)
- `runtime/syscall-server/syscall_context.cpp` — `863b94ff` (#639, 2026-08-13) and `249d987d` (#640, 2026-08-13)

So the Frida map_files/VMA split and the x86 `pt_regs` fixes are not what has to be resolved here; those three commits are.

The `bpf_attach_ctx.cpp` conflict has a trap worth flagging, because it reads as pure formatting. Both sides are the `load_prog_and_helpers` signature at line 40:

- merge base `1dd02f76`: `static int load_prog_and_helpers(bpftime_prog *prog, const runtime_config &config)` — 82 columns
- branch `e1c34844`: the same line with `static` kept, wrapped onto two lines (`.clang-format` sets `ColumnLimit: 80`)
- master `b87f9682`: `int load_prog_and_helpers(bpftime_prog *prog, const runtime_config &config)` — 75 columns, fits

So the reflex is "mine is only a rewrap, keep it". That will not compile. `aa5be1f8` dropped `static` and in the same commit added

- `int load_prog_and_helpers(bpftime_prog *prog, const runtime_config &config);` at `runtime/include/bpf_attach_ctx.hpp:261`, and
- an external caller at `runtime/src/bpf_helper.cpp:564` (`bpftime::load_prog_and_helpers(`)

Both of those merge cleanly, so neither is visible inside the conflict — but `bpf_attach_ctx.cpp` includes that header at line 21, so a `static` definition after the extern declaration is `error: ... was declared 'extern' and later 'static' [-fpermissive]`, and the tree builds C++20 with no `-fpermissive` anywhere.

Take master's side of that hunk. It costs this branch nothing: the branch's only edit there is the line wrap, and master's line already fits inside 80 columns.

For the `syscall_context.cpp` conflict I am only reporting provenance (`863b94ff` and `249d987d`), not a resolution — I did not verify one.

Caveat: the compiler diagnostic above was reproduced on a faithful minimal case, not on a full bpftime build.

PR #367, opened 2024-12-17
feat:Prevent segmentation fault in probe_read & probe_write

Sy0307:
## Description

We avoid core dump when accessing an illegal address (such as NULL). Now we will get different error code for such situation.

## Type of change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:23.04
+FROM ubuntu:24.04
 WORKDIR /bpftime
 RUN apt-get update && apt-get install -y --no-install-recommends \
         libelf1 libelf-dev zlib1g-dev make cmake git libboost1.74-all-dev \


runtime/src/bpf_helper.cpp
```diff
@@ -64,20 +66,93 @@ long bpftime_strncmp(const char *s1, uint64_t s1_sz, const char *s2)
...
+		return -EINVAL;
+	}
+	struct sigaction sa, old_sa;
+	uint64_t ret = 0;
+
+	sa.sa_handler = segv_handler; // set signal handler
+	sigemptyset(&sa.sa_mask); // clear signal set
+	sa.sa_flags = 0;
+
+	if (sigaction(SIGSEGV, &sa, &old_sa) == -1) {
```
Officeyutong, line 90: Do we have any method to optimize these syscalls? For now, each call to bpftime_probe_read may lead to a series syscalls, which are relatively slow. 

Maybe we can setup signal handlers only once, and in signal handler, walk stack trace to dertermine whether a call to bpftime_probe_read triggers this signal, and determine to let the call fail or crash.
Will this be too hard?

runtime/src/bpf_helper.cpp
```diff
@@ -64,20 +66,93 @@ long bpftime_strncmp(const char *s1, uint64_t s1_sz, const char *s2)
...
+		       (size_t)(uint32_t)(size));
+		SPDLOG_TRACE("probe_read: dst={}, src={}, len={}", dst, ptr,
+			     size);
+	} else {
+		SPDLOG_ERROR("probe_read: failed to read from src={}", ptr);
+		ret = -EFAULT;
+	}
+
+	// restore the origin signal handler
+	if (sigaction(SIGSEGV, &old_sa, NULL) == -1) {
```
Officeyutong, line 108: Use `nullptr` in C++, don't use `NULL`

runtime/unit-test/test_probe.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+extern "C" {
+
+uint64_t bpftime_probe_read(uint64_t dst, uint64_t size, uint64_t ptr, uint64_t,
+			    uint64_t);
+uint64_t bpftime_probe_write_user(uint64_t dst, uint64_t src, uint64_t len,
+				  uint64_t, uint64_t);
+
+// prepare for future use
+long bpftime_strncmp(const char *s1, uint64_t s1_sz, const char *s2);
```
Officeyutong, line 14: Maybe you don't need to copy these unused declarations now, I'll sync my fork with your once yours is merged, because my PR still has a long way to go

Officeyutong (changes requested):

runtime/src/bpf_helper.cpp
Sy0307: I will make a try for this idea later.

yunwei37: You can try to use https://github.com/eunomia-bpf/bpftime/blob/master/benchmark%2FREADME.md to see how it performs.

PR #368, opened 2024-12-21
feat:Prevent segmentation fault in probe_read & probe_write

Sy0307:
## Description

We avoid core dump when accessing an illegal address (such as NULL). Now we will get different error code for such situation.

Fixes # (issue)

## Type of change

### DIFF
--- a/.github/workflows/test-runtime.yml
+++ b/.github/workflows/test-runtime.yml
@@ -40,7 +40,7 @@ jobs:
   
     - name: Build
       run: |
-        cmake -DTEST_LCOV=ON -DBPFTIME_LLVM_JIT=YES -DBPFTIME_ENABLE_UNIT_TESTING=YES -DCMAKE_BUILD_TYPE=Debug -B build
+        cmake -DTEST_LCOV=ON -DBPFTIME_LLVM_JIT=YES -DBPFTIME_ENABLE_UNIT_TESTING=YES -DENABLE_PROBE_WRITE_CHECK=1 -DENABLE_PROBE_READ_CHECK=1 -DCMAKE_BUILD_TYPE=Debug -B build
         cmake --build build --config Debug --target bpftime_runtime_tests -j$(nproc)
     - name: Test Runtime
       run:  ./build/runtime/unit-test/bpftime_runtime_tests
--- a/.github/workflows/test-runtime.yml
+++ b/.github/workflows/test-runtime.yml
@@ -69,7 +69,7 @@ jobs:
     - name: build runtime with mpk enable
       run: |
         rm -rf build
-        cmake -Bbuild -DTEST_LCOV=ON -DBPFTIME_LLVM_JIT=YES  -DBPFTIME_ENABLE_UNIT_TESTING=YES -DBPFTIME_ENABLE_MPK=YES -DCMAKE_BUILD_TYPE=Debug
+        cmake -Bbuild -DTEST_LCOV=ON -DBPFTIME_LLVM_JIT=YES  -DBPFTIME_ENABLE_UNIT_TESTING=YES -DBPFTIME_ENABLE_MPK=YES -DCMAKE_BUILD_TYPE=Debug -DENABLE_PROBE_WRITE_CHECK=1 -DENABLE_PROBE_READ_CHECK=1
         cmake --build build --config Debug --target bpftime_runtime_tests -j$(nproc)
     - name: test runtime with mpk
       run:  ./build/runtime/unit-test/bpftime_runtime_tests
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,15 +86,6 @@ endif()
 
 message(STATUS "Started CMake for ${PROJECT_NAME} v${PROJECT_VERSION}...\n")
 
-# option to use probe read/write checks
-if(DEFINED ENABLE_PROBE_READ_CHECK)
-  add_compile_definitions(ENABLE_PROBE_READ_CHECK=1)
-endif()
-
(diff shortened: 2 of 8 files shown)


runtime/src/bpf_helper.cpp
```diff
@@ -64,20 +67,151 @@ long bpftime_strncmp(const char *s1, uint64_t s1_sz, const char *s2)
...
-			    uint64_t)
+extern void jump_point_read();
+extern void jump_point_write();
+/*
+status instruction for probe_read and probe_write
+ -1 = not running
+ 0 = running but no error
+ 1 = running with error
+*/
+thread_local int status_probe_write = -1;
```
Officeyutong, line 78: An enum is better, also , it should be static

Officeyutong (changes requested):

runtime/src/bpf_helper.cpp
```diff
@@ -64,20 +67,158 @@ long bpftime_strncmp(const char *s1, uint64_t s1_sz, const char *s2)
...
+			origin_handler_exist_read = ORIGIN_HANDLER_EXIST;
+			origin_segv_read_handler = original_sa.sa_sigaction;
+		}
+	}
+
+	if (original_sa.sa_sigaction != segv_read_handler) {
+		sa.sa_flags = SA_SIGINFO;
+		sigemptyset(&sa.sa_mask);
+		sa.sa_sigaction = segv_read_handler;
+		sigaction(SIGSEGV, &sa, NULL);
```
Officeyutong, line 142: Please check the return value of syscall. Also use nullptr instead of NULL

Officeyutong (changes requested):

runtime/src/bpf_helper.cpp
```diff
@@ -64,20 +67,186 @@ long bpftime_strncmp(const char *s1, uint64_t s1_sz, const char *s2)
...
 
-uint64_t bpftime_probe_read(uint64_t dst, uint64_t size, uint64_t ptr, uint64_t,
-			    uint64_t)
+extern void jump_point_read();
+extern void jump_point_write();
+
+/*
+status instruction for probe_read and probe_write
+*/
+enum PROBE_STATUS {
```
Officeyutong, line 76: I think you may use enum class, to avoid redundant `PROBE_` in variant names

.github/workflows/test-runtime.yml
```diff
@@ -40,7 +40,7 @@ jobs:
   
     - name: Build
       run: |
-        cmake -DTEST_LCOV=ON -DBPFTIME_LLVM_JIT=YES -DBPFTIME_ENABLE_UNIT_TESTING=YES -DCMAKE_BUILD_TYPE=Debug -B build
+        cmake -DTEST_LCOV=ON -DBPFTIME_LLVM_JIT=YES -DBPFTIME_ENABLE_UNIT_TESTING=YES -DENABLE_PROBE_WRITE_CHECK=1 -DENABLE_PROBE_READ_CHECK=1 -DCMAKE_BUILD_TYPE=Debug -B build
```
Officeyutong, line 43: Remember to add them in the build doc, https://github.com/eunomia-bpf/eunomia.dev/blob/main/docs/bpftime/documents/build-and-test.md

CMakeLists.txt
```diff
@@ -86,6 +86,15 @@ endif()
 
 message(STATUS "Started CMake for ${PROJECT_NAME} v${PROJECT_VERSION}...\n")
 
+# option to use probe read/write checks
+if(DEFINED ENABLE_PROBE_READ_CHECK)
+  add_compile_definitions(ENABLE_PROBE_READ_CHECK=1)
```
Officeyutong, line 91: This definition is only needed for target `runtime`, you may use target_compile_definitions

cmake/StandardSettings.cmake
```diff
@@ -87,4 +87,10 @@ option(BPFTIME_BUILD_KERNEL_BPF "Whether to build with bpf share maps" ON)
 option(BPFTIME_BUILD_STATIC_LIB "Whether to build a single static library for different archive files" OFF)
 
 # whether to build bpftime with libbpf and other linux headers
-option(BPFTIME_BUILD_WITH_LIBBPF "Whether to build with libbpf and other linux headers" ON)
\ No newline at end of file
+option(BPFTIME_BUILD_WITH_LIBBPF "Whether to build with libbpf and other linux headers" ON)
+
+# whether to enable probe write check
+option(ENABLE_PROBE_WRITE_CHECK "Whether to enable probe write check" OFF)
```
Officeyutong, line 93: I suggest it should be enabled by default

runtime/src/bpf_helper.cpp
```diff
@@ -64,20 +69,204 @@ long bpftime_strncmp(const char *s1, uint64_t s1_sz, const char *s2)
...
+/*
+status instruction for probe_read and probe_write
+*/
+enum class PROBE_STATUS {
+	NOT_RUNNING = -1,
+	RUNNING_NO_ERROR = 0,
+	RUNNING_ERROR = 1
+};
+
+thread_local static PROBE_STATUS status_probe_write = PROBE_STATUS::NOT_RUNNING;
```
Officeyutong, line 84: These variables should also be disabled if PROBE_READ/WRITE_CHECK was disabled

cmake/StandardSettings.cmake
Sy0307: But the performance will be poor for default situation.

Officeyutong (approved):

Officeyutong:
@yunwei37

yunwei37 (approved):

PR #371, opened 2025-02-16
fix: use boost_shm_string when an agent_config being setup on shared memory

Officeyutong:
In the previous implementation, field `logger_output_path` in `agent_config` was in type `std::string`, but this struct will be put on shared memory when setting `BPFTIME_LOG_OUTPUT` on server side. This will crash any agent who tries to use configuration from shared memory.

This PR updates it, making it use `boost_shm_string` when the config struct was transfered to shared memory.

### DIFF
--- a/.github/workflows/test-attach.yml
+++ b/.github/workflows/test-attach.yml
@@ -93,6 +93,7 @@ jobs:
           path: |
             ./coverage-syscall-trace.info
       - uses: codecov/codecov-action@v4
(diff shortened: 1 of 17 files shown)


Officeyutong:
> I'm facing some new issue which cause SIGSEGV, and it seems happpens in the code modified by this PR.
> 
> I did some fix by changing it to raw char buffer in #382, could you take a look?

How to reproduce the SIGSEGV?

PR #394, opened 2025-05-05
Implement `bpf_get_stackid`

Officeyutong:
This PR implements `bpf_get_stack` and `bpf_get_stackid`, by using FridaGum to retrive call stacks. 
- Adds map type `BPF_MAP_TYPE_STACK_TRACE`, which is a variant of hash map, it can be used to get stack trace through `bpf_get_stackid`
- Adds function `virtual void *call_attach_specific_function(std::string name, void *data)` on `base_attach_impl`, to call attach-specified functions. With this runtime can call frida_uprobe_attach_impl without depending on it

### DIFF
--- a/.github/workflows/test-examples.yml
+++ b/.github/workflows/test-examples.yml
@@ -202,6 +202,12 @@ jobs:
             syscall_trace: false
             expected_str: "bpf:"
             name: usdt_minimal
+          - path: get_stack_id_example
+            executable: ./get_stack_id_example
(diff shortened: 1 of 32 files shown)


Officeyutong:
> Why a lot of code is not covered by codecov?

I don't know, in theory they should be covered by unit tests and integrated tests, do you have any idea?

yunwei37:
no worries, we can try to improvve it later.

PR #400, opened 2025-05-25
Merge `eGPU` into bpftime

Officeyutong:
**WIP**


Officeyutong:
> Some questions:
> 
> 1. Why update the spdlog and bpftool submodule?
> 2. Maybe it would be easier and clear if you can put the examples and artifacts dir in next pr, not this one?

1. May be a mis-operation, I'll revert it
2. It's ok

example/cuda-probe-retprobe-test/victim
yunwei37, line 1: remove binary plz

tools/cli/main.cpp
```diff
@@ -320,31 +317,32 @@ int main(int argc, const char **argv)
 		} else {
 			return inject_by_frida(pid, agent_path.c_str(), "");
 		}
-	} else if (program.is_subcommand_used("detach")) {
```
yunwei37, line 323: why this is comment out?

attach/nv_attach_impl/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+include(../../cmake/cuda.cmake)
+
+add_library(
+    bpftime_nv_attach_impl STATIC
+    nv_attach_impl.cpp
+    nv_attach_private_data.cpp
+    # cuda_injector.cpp
```
yunwei37, line 7: do we need to use this file?

Officeyutong: `cuda_injector.cpp` is a file written by @vickiegpt and intended for dynamic gpu attach, we may remove it now since we didn't have support for

PR #430, opened 2025-08-13
fix a race condition in userspace ringbuf reservation

Officeyutong:
In the previous implementation, reservation of ringbuf is controlled by a shared lock, which makes multiple threads have a chance to reserve the same memory space. This PR fixes it, uses atomic operations to allocate memory space for each reservation. Thanks for help from @nalreddy

### DIFF
--- a/runtime/src/bpf_map/userspace/ringbuf_map.cpp
+++ b/runtime/src/bpf_map/userspace/ringbuf_map.cpp
@@ -254,9 +254,13 @@ void *ringbuf::reserve(size_t size, int self_fd)
 			size, self_fd);
 		return nullptr;
 	}
-	sharable_lock<interprocess_sharable_mutex> guard(*reserve_mutex);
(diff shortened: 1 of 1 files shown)


runtime/src/bpf_map/userspace/ringbuf_map.cpp
```diff
@@ -244,13 +244,20 @@ struct ringbuf_hdr {
 
 struct spin_lock_guard {
 	pthread_spinlock_t &lock;
-	spin_lock_guard(pthread_spinlock_t &lock) : lock(lock)
+	bool locked;
```
Officeyutong: Why we need this variable?

yunwei37: I think we can remove it...ai is stupid

Officeyutong: removed

PR #442, opened 2025-08-30
fix: Mocked `epoll_wait` doesn't support timeout = -1

### LINKED ISSUE #441: [BUG] Mocked `epoll_wait` doesn't support timeout = -1
In the current implementation, pass `timeout = -1` when calling `epoll_wait` will cause `epoll_wait` to return immediately, without returning anything. But in syscall spec, passing `timeout = -1` should let `epoll_wait` wait indefinitely.

https://github.com/eunomia-bpf/bpftime/blob/e4d6747853a38455a2bec5110dd4fbb383a1cfb1/runtime/src/bpftime_shm.cpp#L397

Officeyutong:
Closes #441 

For details, refer to #441

### DIFF
--- a/runtime/src/bpftime_shm.cpp
+++ b/runtime/src/bpftime_shm.cpp
@@ -360,6 +360,12 @@ int bpftime_is_epoll_handler(int fd)
 int bpftime_epoll_wait(int fd, struct epoll_event *out_evts, int max_evt,
 		       int timeout)
 {
+	if (timeout < 0 && timeout != -1) {
+		SPDLOG_ERROR(
+			"bpftime_epoll_wait only accepts timeout=-1 when negative");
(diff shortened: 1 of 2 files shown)


yunwei37 (approved):

runtime/src/bpftime_shm.cpp
```diff
@@ -360,6 +360,12 @@ int bpftime_is_epoll_handler(int fd)
 int bpftime_epoll_wait(int fd, struct epoll_event *out_evts, int max_evt,
 		       int timeout)
 {
+	if (timeout < 0 && timeout != -1) {
```
yunwei37: timeout < -1?

Officeyutong: fixed

PR #448, opened 2025-09-11
[WIP] Force link VM compat libraries to avoid runtime error in openeuler

### LINKED ISSUE #447: [BUG] Two VMs (ubpf and llvm) can't co-exist on OpenEuler
If compiled with `-DBPFTIME_LLVM_JIT=YES -DBPFTIME_UBPF_JIT=YES`, only LLVM is usable. Setting `BPFTIME_VM_NAME` to `ubpf` will cause a VM not found error

Sy0307:
**WIP PR**

Fix #447 

## Root Cause 

VM factories are registered via global constructors in compat_ubpf.cpp/compat_llvm.cpp.
With both LLVM and uBPF enabled, if final ELF (.so/.exe) doesn’t explicitly reference objects from bpftime_ubpf_vm.a, the linker (default --as-needed/--gc-sections) discards them. The constructor never runs, so "ubpf" isn’t registered, causing:
“No VM factory registered for name: ubpf” / “Unknown VM type requested: ubpf”

## How to fix
~~Force-include bpftime_ubpf_vm/bpftime_llvm_vm at final link of bpftime-agent, bpftime-syscall-server, and bpftime CLI using:~~
~~-Wl,--whole-archive $<TARGET_FILE:bpftime_ubpf_vm> -Wl,--no-whole-archive (same for LLVM)~~

Convert vm/compat/ubpf-vm and vm/compat/llvm-vm to OBJECT-based builds and fold the objects into their static libs to ensure TU with constructors are linked.

Result: objects with registration code are guaranteed in the final ELF; constructors run; "ubpf" is registered and usable at runtime.

**Update:**

After fixing the issue below, we still found some issus, more information in this PR: https://github.com/eunomia-bpf/llvmbpf/pull/34


yunwei37 (approved):

yunwei37:
Hi @Sy0307 , is it completed?

.github/workflows/test-runtime.yml
```diff
@@ -17,28 +17,59 @@ jobs:
     runs-on: "ubuntu-latest"
     strategy:
       matrix:
-        container:
-          - ubuntu-2204
-          - fedora-39
+        container: ["ubuntu-2204", "fedora-39", "openeuler-2403sp2"]
     container:
-      image: "manjusakalza/bpftime-base-image:${{matrix.container}}"
+      image: ${{ fromJSON('{"ubuntu-2204":"manjusakalza/bpftime-base-image:ubuntu-2204","fedora-39":"manjusakalza/bpftime-base-image:fedora-39","openeuler-2403sp2":"openeuler/openeuler:24.03-lts-sp1"}')[matrix.container] }}
```
Officeyutong: We can write `image` field in matrix, instead of extracting from JSON here

.github/workflows/test-runtime.yml
```diff
@@ -50,21 +81,131 @@ jobs:
           dnf install -y dnf-plugins-core
           dnf config-manager --add-repo https://developer.download.nvidia.com/compute/cuda/repos/fedora39/x86_64/cuda-fedora39.repo
           dnf install -y lcov cuda clang-devel
+      - name: Install build tools (Ubuntu)
```
Officeyutong: These empty steps are not needed

Officeyutong (changes requested):

Officeyutong (approved):

PR #466, opened 2025-10-18
Fix error of cpu probe segment fault in gpu shared map

Sy0307:
The previous implementation did not guarantee this, which could cause the agent to hang or fail when attempting to map the IPC handle. Logic has been added to ensure that a valid CUDA context is initialized in the agent process *before* the IPC handle is opened, stabilizing the memory mapping process.

Update readme by english.


yunwei37 (approved):
How much overhead will this bring to server side GPU map update? can you measure it?

Besides that, I think it's a good fix, we need them.

Issue #471, opened 2025-10-18
[FEATURE] Add benchmark for directly run on GPU


yunwei37:
in https://github.com/eunomia-bpf/bpftime/tree/master/example/gpu/directly_run_on_gpu

Can you write a GEMM kernel or vec add kernel in eBPF?

- add a bench for GEMM kernel, compare with raw cuda version
- add a bench for vec add kernel, compare with raw cuda version

PR #473, opened 2025-10-20
[~] feat: Add direct run gpu example for gemm& vec_add

Sy0307:
Solve #471


benchmark/gpu/micro/micro_gemm_config.json
```diff
@@ -19,6 +19,21 @@
     "minimal": "benchmark/gpu/workload/matrixMul 32 3 32 1"
   },
   "test_cases": [
+    {
```
yunwei37, line 22: I think there is no need to add into this framwork and json, you can just run the base scripts?

yunwei37 (approved):

Issue #488, opened 2025-11-03
Unable to map CUDA IPC memory for shared array map, error=201


cizzoo:
I am attempting to run the `gpu/mem_trace` example from the bpftime repository. I am running the components directly on my host machine (not within a Docker container) using root permissions and `CUDA 12.6`.

I am encountering a fatal error in the server component (mem_trace) immediately after it starts.
```
[2025-11-03 15:02:47][info][2617448] Initializing nv_gpu_shared_array_map_impl at pid 2617448
[2025-11-03 15:02:47][error][2617448] Unable to map CUDA IPC memory for shared array map, error=201
[2025-11-03 15:02:47][error][2617448] Exception happened when performing map lookup elem: Unable to map CUDA IPC memory for shared array map!
```

I'm running the example using the following commands, both with root permission:
```bash
BPFTIME_LOG_OUTPUT=console LD_PRELOAD=build/runtime/syscall-server/libbpftime-syscall-server.so CUDA_VISIBLE_DEVICES=1 PATH=/opt/cuda-12.6:$PATH \
  example/gpu/mem_trace/mem_trace
```

and then the agent:

```bash
BPFTIME_LOG_OUTPUT=console LD_PRELOAD=build/runtime/agent/libbpftime-agent.so CUDA_VISIBLE_DEVICES=1 PATH=/opt/cuda-12.6:$PATH example/gpu/mem_trace/vec_add
````

The vec_add application appears to start correctly.
<details>
<summary>agent output</summary>

```
[2025-11-03 15:30:39.402] [info] [bpftime_shm_internal.cpp:927] Registered shared memory with CUDA: addr=0x7feba5343000 size=20971520
[2025-11-03 15:30:39.402] [info] [bpftime_shm_internal.cpp:755] Global shm constructed. shm_open_type 1 for bpftime_maps_shm
[2025-11-03 15:30:39.402] [info] [bpftime_shm_internal.cpp:42] Global shm initialized
[2025-11-03 15:30:39][info][2628804] Initializing CUDA shared memory
[2025-11-03 15:30:41][info][2628804] CUDA context created
[2025-11-03 15:30:41][info][2628804] bpf_attach_ctx constructed
[2025-11-03 15:30:41][info][2628817] CUDA watcher thread started
[2025-11-03 15:30:41][info][2628804] Register attach-impl defined helper bpf_get_func_arg, index 183
[...]

Issue #486, opened 2025-11-01
[BUG] BPF map consistency between host and GPU devices with multiple thread blocks


Forsworns:
**Describe the bug**
I tested two cases, and found the BPF map seems not consistent between host and GPU devices with multiple thread blocks. I'm not sure if they are expected.

**To Reproduce**

Take the  cuda-counter as an example. I tested two cases:

1. By simply revising the grid/block config of `vectorAdd` kernel in 

https://github.com/eunomia-bpf/bpftime/blob/d561e85c1650f932d5a56a49cc4108019eefa51d/example/gpu/cuda-counter/vec_add.cu#L67

to `vectorAdd<<<2, 10>>>(d_A, d_B, d_C);`

The output of the host side becomes weird. The counter should have been increased by 20 every second.

<img width="345" height="189" alt="Image" src="https://github.com/user-attachments/assets/68207191-b635-435b-8b3b-b1a60c759ae1" />

2. Revise the 

https://github.com/eunomia-bpf/bpftime/blob/master/example/gpu/cuda-counter/cuda_probe.c 

to prefill the BPF map before attached.

```c
// ……
int main(int argc, char **argv)
{
	// ……

	/* Load & verify BPF programs */
	err = cuda_probe_bpf__load(skel);
	if (err) {
		fprintf(stderr, "Failed to load and verify BPF skeleton\n");
		goto cleanup;
	}

++>  // Prefill the BPF map before attached.
++>	int fd = bpf_map__fd(skel->maps.call_count);
++>	uint32_t pid = 0;
++>	uint64_t zero = 0;
++>	bpf_map_update_elem(fd, &pid, &zero, BPF_NOEXIST);

	err = cuda_probe_bpf__attach(skel);
	if (err) {
		fprintf(stderr, "Failed to attach BPF skeleton\n");
		goto cleanup;
	}
	while (!exiting) {
		sleep(1);
		print_stat(skel);
	}
cleanup:
	/* Clean up */
	cuda_probe_bpf__destroy(skel);

	return err < 0 ? -err : 0;
}
```

And then fix the BPF map key to 0.

```C
SEC("kprobe/_Z9vectorAddPKfS0_Pf")
int cuda__probe()
{
++>  // Fix the BPF map key to 0
++>	u32 pid = 0;
	u64 ts = bpf_get_globaltimer();
	u64 x, y, z;
	bpf_get_block_idx(&x, &y, &z);
	// Store entry timestamp
[...]

PR #489, opened 2025-11-03
[Bug Fix] Fix SHM open_type judgement

Sy0307:
To solve Issues #488 #486 

From source code we get:
```
bpftime_initialize_global_shm(shm_open_type::SHM_REMOVE_AND_CREATE);
```


runtime/src/bpf_map/gpu/nv_gpu_ringbuf_map.cpp
```diff
@@ -25,7 +25,7 @@ CUdeviceptr
 nv_gpu_ringbuf_map_impl::try_initialize_for_agent_and_get_mapped_address()
 {
 	if (shm_holder.global_shared_memory.get_open_type() !=
-	    shm_open_type::SHM_CREATE_OR_OPEN) {
+	    shm_open_type::SHM_REMOVE_AND_CREATE) {
```
yunwei37, line 28: Why not always SHM_OPEN_ONLY?

Issue #487, opened 2025-11-01
[FEATURE] Add CI for running GPU related examples


Officeyutong:
We have a GPU runner (named `mnfe-nas`) now, add CI to run gpu examples on it

PR #490, opened 2025-11-03
[CI] Add gpu examples CI tests & Fix errors in gpu_example

Sy0307:
To solve #487

### DIFF
--- a/runtime/unit-test/maps/test_shm_hash_maps.cpp
+++ b/runtime/unit-test/maps/test_shm_hash_maps.cpp
@@ -9,7 +9,6 @@
 #include <boost/interprocess/creation_tags.hpp>
 #include <cstdint>
 #include <linux/bpf.h>
-#include <memory>
 #include <sys/wait.h>
 #include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


attach/nv_attach_impl/nv_attach_utils.cpp
```diff
@@ -10,13 +11,40 @@ namespace bpftime
...
 {
 	return TRAMPOLINE_PTX;
 }
 std::string wrap_ptx_with_trampoline(std::string input)
 {
-	return get_defaul_trampoline_ptx() + input;
+	return get_default_trampoline_ptx() + input;
+}
+
+static std::string rewrite_ptx_target(std::string ptx,
```
Officeyutong, line 23: What does this function do?

runtime/src/attach/bpf_attach_ctx_cuda.cpp
```diff
@@ -134,7 +134,51 @@ void bpf_attach_ctx::start_cuda_watcher_thread()
 							     .map_lookup;
 					auto ptr = bpftime_map_lookup_elem(
 						map_fd, req.key);
-					resp.value = ptr;
+					resp.value = nullptr;
+					if (ptr) {
+						const auto host_ptr =
```
Officeyutong, line 139: What does this piece of code do?

runtime/src/bpf_map/gpu_kernel_shared/array_map_kernel_gpu.hpp
```diff
@@ -28,15 +28,17 @@ class array_map_kernel_gpu_impl {
 	uint32_t _max_entries;
 	int map_fd = -1;
 	int kernel_map_id = -1;
+	bool has_kernel_map = false;
```
Officeyutong, line 31: What does this flag mean?

runtime/src/handler/map_handler.cpp
```diff
@@ -727,18 +728,35 @@ int bpf_map_handler::map_init(managed_shared_memory &memory)
 	static CUdevice device;
 #endif
 
+	if (map_impl_ptr.get() != nullptr) {
+		return 0;
+	}
+
+	const auto init_refcnt = [&]() {
```
Officeyutong, line 735: Why do we need counting reference to a map handler? Also, will recording reference locally cause any issues to shared memory?

runtime/syscall-server/syscall_server_utils.cpp
```diff
@@ -24,7 +25,7 @@
 #endif
 namespace bpftime
 {
-static bool already_setup = false;
+static std::once_flag g_startup_once;
```
Officeyutong, line 28: Does `std::once_flag` requires call a an initializer function? If so, will it work correctly if bpftime was loaded before initialization of global variables?

attach/nv_attach_impl/nv_attach_utils.cpp
Sy0307: To support `sm_51`. Let's implement this as a PTX pass later.

runtime/src/attach/bpf_attach_ctx_cuda.cpp
Sy0307: In the trampoline, `_bpf_helper_ext_0001` ultimately returns `resp.map_lookup.value` as a pointer for the GPU code to use (it will subsequently be dereferenced/read/written by the GPU). Previously, `bpftime_map_lookup_elem()` on the syscall-server side returned a pointer within the host process address space. The GPU treated this value as a device pointer, which caused errors like "an illegal memory access was encountered".

runtime/src/bpf_map/gpu_kernel_shared/array_map_kernel_gpu.hpp
Sy0307: This is to fix the error where `kernel_bpf_map_id<=0` for `BPF_MAP_TYPE_GPU_KERNEL_SHARED_ARRAY_MAP`.
However, I just discovered the `BPFTIME_BUILD_WITH_LIBBPF` macro definition. I wonder if this will solve the problem?

runtime/src/handler/map_handler.cpp
Sy0307: `bpftime_shm::dup_bpf_map` here has different semantics from POSIX `dup`. The `dup`ed handler does not construct a new container anymore; instead, it directly reuses the original handler's `map_impl_ptr`, but the corresponding reference count needs to be incremented. This fix is to resolve the timing issue with `directly_on_gpu`.

runtime/syscall-server/syscall_server_utils.cpp
Sy0307: Not needed. This was written to address thread safety issues. After looking into it, it shouldn't be an issue?

runtime/src/bpf_map/gpu_kernel_shared/array_map_kernel_gpu.hpp
Officeyutong: > kernel_bpf_map_id

kernel_bpf_map_id will never be less than 0, since constructing a array_map_kernel_gpu requires a kernel map

runtime/src/attach/bpf_attach_ctx_cuda.cpp
Officeyutong: > In the trampoline, `_bpf_helper_ext_0001` ultimately returns `resp.map_lookup.value` as a pointer for the GPU code to use (it will subsequently be dereferenced/read/written by the GPU). Previously, `bpftime_map_lookup_elem()` on the syscall-server side returned a pointer within the host process address space. The GPU treated this value as a device pointer, which caused errors like "an illegal memory access was encountered".

I remembered I have mapped the whole segment of boost shared memory to GPU, so it's ok to access bytes in shared memory on GPU. Did you encounter any issues?

Sy0307: Error code is 700 and log showed "an illegal memory access was encountered".

runtime/src/bpf_map/gpu_kernel_shared/array_map_kernel_gpu.hpp
Sy0307: `kernel_bpf_map_id` = 0 is possible.

Officeyutong: > `kernel_bpf_map_id` = 0 is possible.

I think 0 is an legal map id, isn't it?

Sy0307: > > `kernel_bpf_map_id` = 0 is possible.
> 
> I think 0 is an legal map id, isn't it?

Yes, but we need to distinguish this from other cases(id >0) to avoid crash.

update: If `kernel_id` is 0, `bpf_map_get_fd_by_id` will return < 0, and subsequently `init_map_fd` inside `try_initialize_for_agent_and_get_mapped_address` will throw an exception.

PR #493, opened 2025-11-05
fix: generate build-time header for PTX pass paths to avoid shell escaping issues

yunwei37:
Build Fix Summary: nv_attach_impl Compilation Issue

  Problem

  The build was failing with undefined reference errors when
  linking bpftimetool and bpftime-aot-cli:

  undefined reference to `bpftime::attach::nv_attach_impl::extract
  _ptxs[abi:cxx11](std::vector<unsigned char, 
  std::allocator<unsigned char> >&&)'
  undefined reference to 
  `bpftime::attach::nv_attach_impl::hack_fatbin(...)`
  undefined reference to
  `bpftime::attach::nv_attach_impl::nv_attach_impl()'

  Root Cause

  The nv_attach_impl.cpp file was failing to compile due to a
  preprocessor macro expansion issue. The
  DEFAULT_PTX_PASS_EXECUTABLE macro was being set to an absolute
  path containing colons (:) as path separators:

  target_compile_definitions(bpftime_nv_attach_impl PUBLIC

  "DEFAULT_PTX_PASS_EXECUTABLE=\"${PTX_PASS_EXECUTABLE_STR}\"")

  This expanded to something like:
  -DDEFAULT_PTX_PASS_EXECUTABLE="/home/user/workspace/bpftime/buil
  d/.../lib1.so:/home/user/workspace/bpftime/build/.../lib2.so"

  When passed through CMake → Make → Shell → Compiler, the complex
   quoting and escaping caused the preprocessor to fail parsing,
  resulting in compilation errors. The object file
  (nv_attach_impl.cpp.o) was only 624 bytes (corrupted/empty),
  causing all symbols to be missing during linking.

  Solution

  Replaced the preprocessor macro approach with a generated header
   file approach:

  Changes to attach/nv_attach_impl/CMakeLists.txt

  Before:
  list(JOIN PTX_PASS_EXECUTABLE ":" PTX_PASS_EXECUTABLE_STR)

  target_compile_definitions(bpftime_nv_attach_impl PUBLIC

  "DEFAULT_PTX_PASS_EXECUTABLE=\"${PTX_PASS_EXECUTABLE_STR}\"")

  After:
  # Create a build-time header file with the PTX pass paths
  # This avoids shell escaping issues with colons and quotes
  file(GENERATE OUTPUT
  "${CMAKE_CURRENT_BINARY_DIR}/ptx_pass_config.h"
       CONTENT "#pragma once\nstatic constexpr const char 
  *DEFAULT_PTX_PASS_EXECUTABLE = 
  \"$<JOIN:${PTX_PASS_EXECUTABLE},:>\";\n")

  target_include_directories(bpftime_nv_attach_impl PRIVATE
  ${CMAKE_CURRENT_BINARY_DIR})

  Changes to attach/nv_attach_impl/nv_attach_impl.cpp

  Added include for the generated header:

  #include "ptxpass/core.hpp"
  #include "ptx_pass_config.h"  // <-- Added this line
  using namespace bpftime;
  using namespace attach;

  Generated Header File

  The CMake configuration now generates
  build/attach/nv_attach_impl/ptx_pass_config.h at configure time:

  #pragma once
  static constexpr const char *DEFAULT_PTX_PASS_EXECUTABLE =
      "/home/yunwei37/workspace/bpftime/build/attach/nv_attach_imp
  l/pass/ptxpass_kprobe_entry/libptxpass_kprobe_entry.so:"
      "/home/yunwei37/workspace/bpftime/build/attach/nv_attach_imp
  l/pass/ptxpass_kretprobe/libptxpass_kretprobe.so:"

  "/home/yunwei37/workspace/bpftime/build/attach/nv_attach_impl/pa
  ss/ptxpass_kprobe_memcapture/libptxpass_kprobe_memcapture.so";

  Benefits

  1. Avoids shell escaping issues: No special characters need to
  be escaped through multiple shell layers
  2. Cleaner approach: Separates configuration from compilation
  3. Uses CMake generator expressions: Properly handles build-time
   path expansion with $<JOIN:...>
  4. Type-safe: Uses C++ constexpr instead of preprocessor macros

  Verification

  - Object file size increased from 624 bytes to 13 MB (properly
  compiled)
  - All symbols are now present:
    - bpftime::attach::nv_attach_impl::extract_ptxs
    - bpftime::attach::nv_attach_impl::hack_fatbin
    - bpftime::attach::nv_attach_impl::nv_attach_impl()
  (constructor)
  - Both bpftimetool and bpftime-aot-cli link successfully
  - Full build completes without errors

  Files Modified

  1. attach/nv_attach_impl/CMakeLists.txt - Replaced
  target_compile_definitions with file(GENERATE ...)
  2. attach/nv_attach_impl/nv_attach_impl.cpp - Added #include 
  "ptx_pass_config.h"

  Files Generated (at build time)

  - build/attach/nv_attach_impl/ptx_pass_config.h - Contains the
  DEFAULT_PTX_PASS_EXECUTABLE constant

### DIFF
--- a/attach/nv_attach_impl/CMakeLists.txt
+++ b/attach/nv_attach_impl/CMakeLists.txt
@@ -39,7 +39,6 @@ set_property(TARGET bpftime_nv_attach_impl PROPERTY CXX_STANDARD 20)
 file(GENERATE OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/ptx_pass_config.h"
      CONTENT "#pragma once\nstatic constexpr const char *DEFAULT_PTX_PASS_EXECUTABLE = \"$<JOIN:${PTX_PASS_EXECUTABLE},:>\";\n")
 
-target_include_directories(bpftime_nv_attach_impl PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
 target_link_options(bpftime_nv_attach_impl PUBLIC "-Wl,--whole-archive" "$<TARGET_FILE:bpftime_nv_attach_impl>" "-Wl,--no-whole-archive")
 


attach/nv_attach_impl/CMakeLists.txt
```diff
@@ -34,9 +34,11 @@ target_link_libraries(bpftime_nv_attach_impl PUBLIC bpftime_base_attach_impl spd
 
 set_property(TARGET bpftime_nv_attach_impl PROPERTY CXX_STANDARD 20)
 
-list(JOIN PTX_PASS_EXECUTABLE ":" PTX_PASS_EXECUTABLE_STR)
+# Create a build-time header file with the PTX pass paths
+# This avoids shell escaping issues with colons and quotes
+file(GENERATE OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/ptx_pass_config.h"
+     CONTENT "#pragma once\nstatic constexpr const char *DEFAULT_PTX_PASS_EXECUTABLE = \"$<JOIN:${PTX_PASS_EXECUTABLE},:>\";\n")
```
Officeyutong, line 40: Do we need `target_include_directories`?

yunwei37: It's already in line 29: target_include_directories(bpftime_nv_attach_impl PRIVATE ${SPDLOG_INCLUDE} ${CMAKE_CURRENT_BINARY_DIR}

PR #495, opened 2025-11-09
[+] Feat: Add cuda_memcpy_to_symbol for fixing constant sync issue 

Sy0307:
To fix issue of example/gpu/lanuchlate

before fix
```
C[1] = 0 (expected 3)
```

after fix

```
C[1] = 3 (expected 3)
```

### DIFF
--- a/attach/nv_attach_impl/nv_attach_impl.cpp
+++ b/attach/nv_attach_impl/nv_attach_impl.cpp
@@ -238,6 +238,20 @@ nv_attach_impl::nv_attach_impl()
 			      cuda_malloc_managed_addr);
 	}
 	{
+		void *cuda_memcpy_to_symbol_addr =
+			GSIZE_TO_POINTER(gum_module_find_export_by_name(
+				nullptr, "cudaMemcpyToSymbol"));
+		register_hook(AttachedToFunction::CudaMemcpyToSymbol,
+			      cuda_memcpy_to_symbol_addr);
+	}
+	{
+		void *cuda_memcpy_to_symbol_async_addr =
+			GSIZE_TO_POINTER(gum_module_find_export_by_name(
+				nullptr, "cudaMemcpyToSymbolAsync"));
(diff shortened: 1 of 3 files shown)


Officeyutong (approved):

Sy0307:
> @Sy0307 I guess this is also related to the constant value assignment problem in #481 ? Could you please share why the API `cudaMemcpyToSymbol` has to be hooked and replaced with customized copy?

Maybe it will solve your issue. The reason is VecAdd must read the right constant value on runtime. Bpftime only used to record symbol mapping when `__cudaRegisterVar` , but we need to trace cudaMemcpyToSymbol to sync symbol.

Forsworns:
@Sy0307  Thanks, but I'm still confused about this...
- Why the original `cudaMemcpyToSymbol ` API from user does not work? Is it due to the just-in-time modification to the fatbin, so that the layout of the fatbin changes and the original symbol address invaid?
- And I found that you use `gum_interceptor_attach` to fix it, instead of `gum_interceptor_replace`. I think the original `cudaMemcpyToSymbol ` will also be called in this way.

Sy0307:
> @Sy0307 Thanks, but I'm still confused about this...
> 
> * Why the original `cudaMemcpyToSymbol ` API from user does not work? Is it due to the just-in-time modification to the fatbin, so that the layout of the fatbin changes and the original symbol address invaid?
> * And I found that you use `gum_interceptor_attach` to fix it, instead of `gum_interceptor_replace`. I think the original `cudaMemcpyToSymbol ` will also be called in this way.

1.  original `cudaMemcpyToSymbol ` will not update __constant__ in bpftime kernel. The data in __constant__ will not make effort.
2. We extract the symbol, src, count, offset, kind, and stream parameters, and then call mirror_cuda_memcpy_to_symbol to write the same data again to the CUdeviceptr. The user still observes standard CUDA behavior, as bpftime performs an extra copy in a side-channel to achieve transparent synchronization.

Any other issue? @Forsworns

PR #501, opened 2025-11-16
[feature] Support for pytorch on GPU

Officeyutong:
This PR adds support for pytorch, including the following performance increase:
- Do not add trampoline for PTXs without probes
- Parallel JIT compile for PTXs
- Cache for patched PTXs, compile PTXs, loading modules (modules loaded with the same PTX will only be loaded once) to reduce memory usage and increase startup speed
- Split PTX compiler into an individual shared library to avoid global variable initialization issue of nvptxjitcompiler
- Use nvptxjitcompiler to compile PTX so the process for compiling PTX can be parallelized

And other stuff:
- A simple example based on threadhist for pytorch

pytorch needs to be recompiled for bpftime. This is my command line to compile it: `USE_NCCL=0 USE_MPI=0 TORCH_CUDA_ARCH_LIST=6.1+PTX USE_XPU=0 USE_ROCM=0 REL_WITH_DEB_INFO=1 CMAKE_POLICY_VERSION_MINIMUM=3.5 CUDA_HOME=/usr/local/cuda-12.6 uv pip install --no-build-isolation -v -e .
`

### DIFF
--- a/example/gpu/launchlate/launchlate.c
+++ b/example/gpu/launchlate/launchlate.c
@@ -108,7 +108,7 @@ int main(int argc, char **argv)
 	struct timespec ts_mono, ts_real;
 	int64_t offset_ns;
 	uint32_t key = 0;
-	const char *uprobe_spec = "example/gpu/launchlate/vec_add:_Z16cudaLaunchKernelIcE9cudaErrorPT_4dim3S3_PPvmP11CUstream_st";
+	const char *uprobe_spec = "./vec_add:_Z16cudaLaunchKernelIcE9cudaErrorPT_4dim3S3_PPvmP11CUstream_st";
 	char *binary_path = NULL;
 	char *func_name = NULL;
 	char *colon_pos = NULL;


attach/nv_attach_impl/nv_attach_fatbin_record.cpp
```diff
@@ -81,55 +102,153 @@ void fatbin_record::try_loading_ptxs(class nv_attach_impl &impl)
...
+					SPDLOG_INFO(
+						"Start compiling {}, not found in cache",
+						name);
+					auto compiler = handler.create();
+					if (!compiler) {
+						throw std::runtime_error(
+							"Unable to create nv_attach_impl_ptx_compiler");
+					}
+					const char *compile_options[] = {
+						"--gpu-name=sm_61", "--verbose",
```
yunwei37: can we make this compile process into a pass or some configurable things?

attach/nv_attach_impl/nv_attach_fatbin_record.cpp
```diff
@@ -81,55 +102,153 @@ void fatbin_record::try_loading_ptxs(class nv_attach_impl &impl)
...
+					compiled_ptx[name] = compiled_program;
+					this->ptx_pool->insert(std::make_pair(
+						sha256_string,
+						compiled_program));
+					SPDLOG_INFO("Compile of {} done", name);
+				}
+			});
+	}
+	pool.join();
+	for (const auto &[name, ptx_and_trampoline_flag] : patched_ptx) {
```
yunwei37, line 196: from a linter view, i would suggest break this very long function into smaller ones...

attach/nv_attach_impl/nv_attach_fatbin_record.cpp
yunwei37: And we have a env to spec the SM version...

example/gpu/threadhist/threadhist.bpf.c
```diff
@@ -23,7 +23,8 @@ int cuda__retprobe()
 {
 	u32 key = 0;
 	u64 *cnt = bpf_map_lookup_elem(&call_count, &key);
-	if (cnt)
+	// bpf_printk("cnt=%lx\n", (unsigned long)cnt);
+	if (cnt && (unsigned long)cnt != 0x7fffbec74400)
```
yunwei37: why?

runtime/include/bpftime_shm.hpp
```diff
@@ -39,7 +39,7 @@ struct bpf_map_attr {
 
 	// additional fields for bpftime only
 	uint32_t kernel_bpf_map_id = 0;
-	uint32_t gpu_thread_count = 1000;
+	uint64_t gpu_thread_count = 1024 * 1024;
```
yunwei37, line 44: What's the thread count here, why we change them? It would be better to have some comments

yunwei37 (approved):

example/gpu/threadhist/threadhist.bpf.c
Officeyutong: It's for debugging, removed now

runtime/include/bpftime_shm.hpp
Officeyutong: Updated comment. pytorch may launch a kernel with 512*1024 threads, so I changed it to this number

attach/nv_attach_impl/nv_attach_fatbin_record.cpp
Officeyutong: It's broke into two functions now

PR #502, opened 2025-11-16
impl builtin GPU hashmap

### LINKED ISSUE #496: [FEATURE] Request BPF HashMap support on GPU

A simple implementation based on https://github.com/eunomia-bpf/bpftime/blob/88012fbfac5eccf75df5f8b6ed70847f11b7e385/runtime/src/bpf_map/bpftime_hash_map.hpp#L14  and https://github.com/eunomia-bpf/bpftime/blob/88012fbfac5eccf75df5f8b6ed70847f11b7e385/runtime/src/bpf_map/gpu/nv_gpu_shared_array_map.hpp#L20

I once tried to add dependency [cuCollections](https://github.com/NVIDIA/cuCollections/tree/dev) to bpftime. In this manner, more GPU BPF Maps can be implemented based on the advanced STL-like containers in cuCollections. But soon I realize that the underlying device memory of cuCollection container cannot be accessed and shared among processes, as the `cuIpcOpenMemHandle` in `nv_gpu_shared_array_map_impl`. Thus, it seems not a good idea to apply cuCollections directly. Or do you have some options to this method?

**Provide usage examples**
E.g., implement a thread-local cache from the GPU side, so that different kernels launched from different threads can access different data. The thread index can be accessed by `bpf_get_current_pid_tgid` and used as the key of BPF hashmap.

Forsworns:
## Description

Build a simple HashMap with keys and indicators stored at the agent host side. It fallbacks all of operations to agent host side for bpf map helpers. It is based on `nv_gpu_shared_array_map_impl`.

It also fixes two hidden problems in the `nv_gpu_shared_array_map_impl` slow path: 
- The `elem_lookup()` and `elem_update()` may get null context since it works in a different thread from the user thread. It was fixed by an additional `cuCtxPushCurrent_v2`, paired with `cuCtxPopCurrent`.
- The `cuMemcpyHtoD` and `cuMemcpyDtoH` used in the `nv_gpu_shared_array_map_impl` implicitly synchronize with the kernel launched by user. It can be solved by an additional `CUstream` at the bpftime agent side.

These issues were hidden by the fast path in trampoline, by removing the fast path in the trampoline. They can be reproduced in the `nv_gpu_shared_array_map_impl` example. 

Fixes #496

## Type of change

## How Has This Been Tested?

Revise the gpu example for shared maps. The vector add example is with pid `80472` (right window), and we get the same value at the server side (left window).

image

**Test Configuration**:

- Hardware: L20
- Toolchain: clang 18.1.3
- SDK: CUDA 12.8

### DIFF
--- a/.github/workflows/test-ptxpass.yml
+++ b/.github/workflows/test-ptxpass.yml
@@ -31,7 +31,7 @@ jobs:
           git config --global --add http.proxy http://192.168.15.1:2345
           git config --global --add https.proxy http://192.168.15.1:2345
           
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
         with:
           submodules: recursive
 


Forsworns:
@yunwei37 @Sy0307 Could you please help review this patch?

Forsworns:
rebase to resolve conflicts

yunwei37:
LGTM, thanks! 👍

Officeyutong:
Please sign your commits, after that I will merge this PR

PR #503, opened 2025-11-17
Fix GPU-CPU memory synchronization and illegal memory access on NVIDI…

dr-xjx:
…A Orin X

This commit addresses critical memory consistency and access issues when running on NVIDIA Tegra platforms (Orin/Thor).

Changes:
1. Add volatile qualifier to CommSharedMem members
   - Mark flag1, flag2, request_id, and map_id as volatile
   - Prevents compiler optimizations that could cause data inconsistency
   - Ensures each access reads from/writes to actual memory

2. Use ld.volatile.global.u32 for flag2 in make_helper_call
   - Forces memory load on each iteration in spin-wait loop
   - Bypasses L1 cache to ensure visibility of CPU writes
   - Critical for cross-device synchronization

3. Migrate to cudaHostAlloc for CommSharedMem allocation
   - Replaces standard library memory allocation (malloc/new)
   - Uses pinned memory to resolve CUDA_ERROR_ILLEGAL_ADDRESS
   - Follows NVIDIA Memory Management guidelines for Tegra platforms
   - Reference: https://docs.nvidia.com/cuda/cuda-for-tegra-appnote/index.html#memory-management

4. Align CommSharedMem lifetime with bpftime_attach_ctx
   - Implement RAII pattern using std::unique_ptr with custom deleter
   - Ensures proper resource cleanup on context destruction

Root Cause:
On NVIDIA Orin/Thor, GPU access to constData allocated via standard library functions triggers illegal memory access errors. The error occurs in _bpf_helper_ext_0006 when executing:
  ...
  ld.const.u64 %rd1, [constData];
  ...
  add.s64 %rd27, %rd1, %rd43;
  st.u8 [%rd27+24], %rs1;

Error: CUDA_ERROR_ILLEGAL_ADDRESS (error code: 700)

Solution:
Using cudaHostAlloc with pinned memory ensures proper memory mapping between CPU and GPU address spaces on integrated GPU systems.

Testing:
- Platform: NVIDIA Orin X
- CPU: ARM Cortex-A78AE (12-core)
- CUDA Driver/Runtime: 12.1 / 11.4
- Compute Capability: 8.7
- Memory: 25767 MBytes
- Test case: example/gpu/cuda-counter
- Status: Verified working

**Please try to use the copilot to summary your PR. You don't need to fill all info below, just it can help giving your a checklist.**

## Description

Fixes # (issue)

## Type of change

## How Has This Been Tested?

- [ ] Test A
example/gpu/cuda-counter
- [ ] Test B

**Test Configuration**:
- Platform: NVIDIA Orin X
- CPU: ARM Cortex-A78AE (12-core)
- CUDA Driver/Runtime: 12.1 / 11.4
- Compute Capability: 8.7
- Memory: 25767 MBytes
- Test case: example/gpu/cuda-counter
- Status: Verified working
- Firmware version:
- Hardware:
- Toolchain:
- SDK:

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [ ] I have checked my code and corrected any misspellings


dr-xjx:
@yunwei37  please review this，tks~

dr-xjx:
@copilot

attach/nv_attach_impl/trampoline/default_trampoline.cu
yunwei37, line 1: Do we need to regenerate the ptx embed in the header file?

runtime/src/attach/bpf_attach_ctx_cuda.cpp
```diff
@@ -77,128 +78,177 @@ bpf_attach_ctx::find_nv_attach_impl() const
...
+
+	CUDAContext::CommSharedMemPtr cuda_shared_mem(ptr);
+
+	SPDLOG_INFO("Allocated {} GB of pinned memory using cudaHostAlloc",
+		    sizeof(CommSharedMem) / (1024.0 * 1024 * 1024));
+
+	memset(cuda_shared_mem.get(), 0, sizeof(CommSharedMem));
+
+	SPDLOG_INFO(
+		"✅ Successfully allocated pinned memory via cudaHostAlloc");
```
yunwei37: (Please remove emoji)

attach/nv_attach_impl/trampoline/default_trampoline.cu
dr-xjx: Should we recompile trampoline_ptx.h on every cmake build?

dr-xjx:
> Thanks a lot for the PR!
> 
> It seems like it fix many bugs in one PR, would you mind seperate them into mmultiple?
> 
> > Use ld.volatile.global.u32 for flag2 in make_helper_call
> 
> Also, does it has performance impact? If yes, how can we improve it while keep it correct?
> 
> @Officeyutong @Sy0307 What do you think?
These changes should be kept as a cohesive whole, and splitting them isn't really necessary.

I believe this PR should have no major impact, but there are minor considerations:

1. **Memory Allocation**: `cudaHostAlloc` requires contiguous physical memory allocation, which may fail when system memory is fragmented or under pressure

2. **Instruction Overhead**: `ld.volatile` is unnecessary for uncached memory access scenarios, but the performance impact is negligible

attach/nv_attach_impl/trampoline/default_trampoline.cu
dr-xjx: recompile trampoline_ptx.h on every cmake build
https://github.com/eunomia-bpf/bpftime/pull/503/commits/46615851cacff75529af910f879addfa54fb1eca

example/gpu/cuda-counter/cuda_probe.bpf.c
```diff
@@ -36,6 +36,7 @@ static const u64 (*bpf_get_thread_idx)(u64 *x, u64 *y, u64 *z) = (void *)505;
 SEC("kprobe/_Z9vectorAddPKfS0_Pf")
 int cuda__probe()
 {
+#if 0
```
Officeyutong: Why comment these sentences out?

dr-xjx: sry,this is debug. accidentally added it

yunwei37:
Thanks! We are testing it locally these days

yunwei37 (changes requested):
Found two blocking issues:

1. In `runtime/src/attach/bpf_attach_ctx_cuda.cpp`, the watcher thread now stores `flag2 = 1` before the host-side fence. The device side treats `flag2` as the completion signal in `attach/nv_attach_impl/trampoline/default_trampoline.cu`, so this reordering allows the GPU to observe completion before the response payload is globally visible. The old code fenced before publishing `flag2`; this version removes that guarantee and can still race on stale `resp` data.

2. `create_cuda_context()` switched to `cudaHostAlloc(..., cudaHostAllocDefault)`, but the communication buffer is later consumed from separately created CUDA contexts (for example `cuCtxCreate()` in `attach/nv_attach_impl/nv_attach_impl.cpp`). Per NVIDIA runtime docs, `cudaHostAllocPortable` is the flag that makes the allocation be treated as pinned across all CUDA contexts. With `Default`, this can still break when allocation and execution happen in different contexts.

Please fix those before merge.

yunwei37 (approved):
Re-reviewed the current implementation and pushed `37ceccd` to close the remaining cross-context registration blocker. The completion flag is already published after the host fence on the current branch; the Boost shared segment is now registered with `cudaHostRegisterMapped | cudaHostRegisterPortable`, matching the later `cuCtxCreate` usage and the existing portable driver-API pattern elsewhere in the runtime.

Validation on the exact head:
- CUDA-enabled `runtime` and `bpftime_runtime_tests` build with CUDA 12.9
- `CUDA watcher stops before shared-memory teardown` passes on an RTX 5090
- `git diff --check`
- dedicated read-only review: PASS
- independent Kimi K3 read-only review: PASS

All 15 review threads are resolved. The earlier `Test cuda compile` local failure is the pre-existing empty VM-factory initialization path and does not exercise shared-memory registration. GitHub CI has restarted; Tegra/Orin hardware confirmation remains external. No merge was performed.

PR #517, opened 2025-11-29
Feat/per thread timing dist

pippaye:
This pull request introduces a new GPU per-thread timing distribution demo under the `example/gpu/per-thread-timing-dist` directory. The demo includes a CUDA kernel that runs a synthetic workload, an eBPF program that collects per-thread timing data, and a user-space application to display timing distributions and summary statistics. The changes provide a complete workflow for measuring and visualizing the runtime distribution of GPU threads using eBPF.

**New Demo Implementation:**

* Added a `Makefile` to automate building the CUDA kernel, eBPF program, and user-space application, including support for third-party dependencies and environment setup.
* Introduced the `timed_work_kernel.cu` CUDA kernel that runs a synthetic workload with varying iterations per thread to generate a non-trivial timing distribution.

**eBPF Program for Timing Collection:**

* Implemented `timing-dist.bpf.c` eBPF program to trace entry/exit of the CUDA kernel, record per-thread start times, and update a log2-based histogram and summary statistics using custom GPU helper functions.

**User-space Application for Visualization:**

* Added `timing-dist.c` user-space program that attaches the eBPF program, periodically prints the timing distribution histogram and summary statistics, and handles graceful termination via signals.

**Project Structure and Build Artifacts:**

* Updated `.gitignore` to exclude build outputs and binaries for the new demo.**Please try to use the copilot to summary your PR. You don't need to fill all info below, just it can help giving your a checklist.**


yunwei37 (changes requested):
Reviewed exact head `6bc62dc97b6f1e65c0ae90ec757a075b4d05da36` and built the new host target in an isolated checkout.

Two correctness/portability issues need fixing:

1. `example/gpu/per-thread-timing-dist/timing-dist.bpf.c:126-134` performs only one CAS for each min/max update and ignores a failed CAS. With all GPU threads contending on the single summary entry, a more-extreme sample can lose the race and never be retried. For example, if samples 80 and 50 both read min=100, 80 can win first and the CAS for 50 then fails, leaving the reported min at 80. Please use a retrying atomic min/max update (stop when the sample no longer improves the current value).

2. `example/gpu/per-thread-timing-dist/timing-dist.c:73-78` passes `1ULL << i` (`unsigned long long`) to `PRIu64`, which expands to `%lu` on Ubuntu/aarch64 and expects `unsigned long`. The target builds, but GCC 13 emits three `-Wformat` warnings and the mismatch is undefined/ non-portable for variadic formatting. Cast the shifted value to `uint64_t` (or otherwise make the argument type match `PRIu64`).

Validation:

- `make -C example/gpu/per-thread-timing-dist timing-dist V=1` succeeded in the Linux Coder environment and emitted the three format warnings above.
- A minimal GCC 13 `-Wall -Wextra -Wformat=2 -fsyntax-only` reproduction on `spark-9773` emitted the same three `%lu` versus `unsigned long long` warnings.
- `git diff --check origin/master...HEAD` passed.

The summary-map initialization is present at `timing-dist.c:112-125`; I explicitly excluded the contrary candidate after checking the complete file.

PR #518, opened 2025-11-29
Implement gpu-kernel shared map

Officeyutong:
This PR adds a map named array_map_kernel_gpu, which can be accessed from both kernel and GPU side
-  kprobe attaches prefixed with `cuda_` will be attached to GPU, otherwise it will be attached as normal kprobe
- The map has type 1504, which isn't compatible with previous kernel-user shared map (kernel-user shared maps uses directly original map types, such as 2(array map)), so I added a translating function to translate original bpftime map ids (such as 1504) to kernel map fds.
- Adds an example to demonstrate this PR. There are two programs in the eBPF program, the uretprobe one will be handled by kernel, and increases index 1 of an shared array when cudaLaunchKernel is called. the kretprobe one will be executed on GPU, and increases index 0 of the shared array. The userspace program keeps printing content of the shared array.

### DIFF
--- a/example/gpu/launchlate-kernel-gpu-shared-map/.gitignore
+++ b/example/gpu/launchlate-kernel-gpu-shared-map/.gitignore
@@ -1,0 +1,7 @@
+/launchlate
+/.output
+/victim*
+/vec_add.cpp
+/vec_add
+/vec_add-new.cpp
+launchlate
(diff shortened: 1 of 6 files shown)


Officeyutong:
More things to do:
- Use prefix (like `cuda|` to distinguish attaches to GPU or to kernel)
- Update launch late to use gpu-kernel shared maps

PR #529, opened 2025-12-14
[+]Feat: Enhance CUDA Graph support and miscs changes 

### LINKED ISSUE #527: [FEATURE] Support cudaGraphLaunch
Support cudaGraph in bpftime

Sy0307:
Fix #527 

Descriptions generated by AI ,reviewed & modified by me.
## 1) CUDA Graph hook

- **Hook Implementation**: Added and refined hook entry points for CUDA Graph-related Driver APIs (e.g., `cuGraphAddKernelNode(_v2)`, `cuGraphKernelNodeSetParams(_v2)`, `cuGraphExecKernelNodeSetParams(_v2)`, etc.) in `attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp`.
- **Kernel Replacement**: Within the parameters of these APIs (`CUDA_KERNEL_NODE_PARAMS_v1/v2`), the target `CUfunction/CUkernel` is now **replaced with the patched CUfunction generated by bpftime**. This ensures that kernels launched repeatedly within a Graph also follow the "patched kernel" execution path.
- **Mapping Logic**: Completed the mapping caching logic for "reverse looking up kernel name via `CUfunction/CUkernel` $\to$ querying `patched_kernel_by_name`," ensuring stable identification and replacement of "original kernel $\to$ patched kernel" in Graph scenarios.

## 2) Fix for CUDA watcher thread exit crash

- In `runtime/src/attach/bpf_attach_ctx_cuda.cpp` and `runtime/src/attach/bpf_attach_ctx.cpp`, ensured that the watcher thread performs a `stop + join` before shared memory is unmapped.
- This is guaranteed via `std::atexit` + `std::call_once` (as a fallback) and within the destructor. This prevents core dumps caused by the watcher thread continuing to read/write CUDA-mapped shared memory that has already been freed or unbound during the exit phase.

## 3) prefault_range

- Added prefault logic (reading page-by-page to trigger page faults) for the shared memory segment in `bpftime_shm::register_cuda_host_memory()` within `runtime/src/bpftime_shm_internal.cpp`.
- **Reasoning**: Shared memory pages allocated by Boost.Interprocess might not yet have Page Table Entries (PTEs) established in the current process. `cudaHostRegister()` only pins pages that are already mapped/faulted-in within the current process. Consequently, if the GPU attempts to access a page at a deeper offset that hasn't been faulted-in, it may hang or fail to return. Prefaulting ensures the entire segment is actually mapped to the process's page table before registration.

## 4) Reason for removing `if (p == ".target" && !target_sm.empty())`

- This logic performed a direct text replacement on the `.target` directive in the PTX header.
- **Issue**: This was prone to creating inconsistencies between header information (like `.version/.target/...`) and the actual PTX content or spliced snippets, subsequently triggering an `nvPTXCompilerCompile()` error code `3` (`Unable to compile: 3`).
- **Solution**: This "text-level patching" was removed in favor of relying on compiler arguments to handle target architecture selection.

### DIFF
--- a/attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp
+++ b/attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp
@@ -354,7 +354,7 @@ cuda_graph_maybe_get_kernel_name_from_cukernel(CUkernel kernel)
 	const char *name = nullptr;
 	if (auto err = cu_kernel_get_name(&name, kernel); err != CUDA_SUCCESS)
 		return std::nullopt;
-	if (name == nullptr || name[0] == '\0')
+	if (name == nullptr)
 		return std::nullopt;
 	return std::string(name);
 }


attach/nv_attach_impl/pass/ptxpass_core/src/core.cpp
```diff
@@ -167,13 +167,13 @@ std::string filter_out_version_headers_ptx(const std::string &input,
...
 				} else {
 					seen.insert(p);
-					// Replace .target with target_sm if provided
-					if (p == ".target" && !target_sm.empty()) {
-						line = ".target " + target_sm;
-						SPDLOG_INFO(
-							"Replaced .target with {}",
-							target_sm);
-					}
+					// Remove for conflict of PTX headers info
```
Officeyutong: If these code pieces are not used anymore, just remove them

attach/nv_attach_impl/nv_attach_impl.cpp
```diff
@@ -352,15 +416,116 @@ nv_attach_impl::nv_attach_impl()
...
+		auto interceptor = (GumInterceptor *)frida_interceptor;
+		gum_interceptor_begin_transaction(interceptor);
+
+		if (frida_listener != nullptr) {
+			gum_interceptor_detach(
+				interceptor,
+				(GumInvocationListener *)frida_listener);
+		}
+
+		static const char *const kReplacedSymbols[] = {
```
Officeyutong: Note that we are using snake case

attach/nv_attach_impl/nv_attach_impl.cpp
```diff
@@ -352,15 +416,116 @@ nv_attach_impl::nv_attach_impl()
...
+			"cuGraphAddKernelNode",
+			"cuGraphAddKernelNode_v2",
+			"cuGraphExecKernelNodeSetParams",
+			"cuGraphExecKernelNodeSetParams_v2",
+			"cuGraphKernelNodeSetParams",
+			"cuGraphKernelNodeSetParams_v2",
+		};
+		revert_frida_replaced_exports_if_present(
+			interceptor, kReplacedSymbols,
+			sizeof(kReplacedSymbols) / sizeof(kReplacedSymbols[0]));
```
Officeyutong: We can use `std::size`

attach/nv_attach_impl/nv_attach_impl.cpp
```diff
@@ -352,15 +416,116 @@ nv_attach_impl::nv_attach_impl()
...
+	}
 	if (ptx_compiler_dl_handle) {
 		dlclose(ptx_compiler_dl_handle);
 	}
 }
+
+void nv_attach_impl::record_patched_kernel_function(const char *kernel_name,
+						    CUfunction function)
+{
+	if (kernel_name == nullptr || kernel_name[0] == '\0' ||
```
Officeyutong: May be we can pass `kernel_name` as a `const std::string&` to avoid such checks..

attach/nv_attach_impl/nv_attach_impl.hpp
```diff
@@ -166,6 +189,9 @@ class nv_attach_impl final : public base_attach_impl {
 		pass_configurations;
 	std::map<std::string, ptxpass::runtime_response::RuntimeResponse>
 		patch_cache;
+	mutable std::mutex cuda_symbol_map_mutex;
```
Officeyutong, line 192: Why do we need the reverse query?

attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp
```diff
@@ -51,6 +54,75 @@ G_DEFINE_TYPE_EXTENDED(
...
+using cu_graph_add_kernel_node_v2_fn_t =
+	CUresult (*)(CUgraphNode *, CUgraph, const CUgraphNode *, size_t,
+		     const CUDA_KERNEL_NODE_PARAMS_v2 *);
+using cu_graph_exec_kernel_node_set_params_v1_fn_t = CUresult (*)(
+	CUgraphExec, CUgraphNode, const CUDA_KERNEL_NODE_PARAMS_v1 *);
+using cu_graph_exec_kernel_node_set_params_v2_fn_t = CUresult (*)(
+	CUgraphExec, CUgraphNode, const CUDA_KERNEL_NODE_PARAMS_v2 *);
+using cu_graph_kernel_node_set_params_v1_fn_t =
+	CUresult (*)(CUgraphNode, const CUDA_KERNEL_NODE_PARAMS_v1 *);
+using cu_graph_kernel_node_set_params_v2_fn_t =
```
Officeyutong, line 67: We can use decltype()

attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp
```diff
@@ -234,38 +312,252 @@ cuda_runtime_function__cudaLaunchKernel(const void *func, dim3 grid_dim,
...
-			cuGetErrorString(err, &error_string);
-			SPDLOG_ERROR("Unable to launch kernel: {} ({})",
-				     error_name ? error_name : "UNKNOWN",
-				     error_string ? error_string :
-						    "No description");
-			SPDLOG_ERROR("Error code: {}", (int)err);
-			return cudaErrorLaunchFailure;
-		}
-		return cudaSuccess;
+	if (impl != nullptr && spdlog::should_log(spdlog::level::debug)) {
```
Officeyutong: Why adding this check (I mean `spdlog:should_log...`)

attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp
```diff
@@ -234,38 +312,252 @@ cuda_runtime_function__cudaLaunchKernel(const void *func, dim3 grid_dim,
...
+{
+	using cu_kernel_get_name_fn_t = CUresult (*)(const char **, CUkernel);
+	static cu_kernel_get_name_fn_t cu_kernel_get_name =
+		(cu_kernel_get_name_fn_t)dlsym(RTLD_DEFAULT, "cuKernelGetName");
+	if (!cu_kernel_get_name)
+		return std::nullopt;
+	const char *name = nullptr;
+	if (auto err = cu_kernel_get_name(&name, kernel); err != CUDA_SUCCESS)
+		return std::nullopt;
+	if (name == nullptr || name[0] == '\0')
```
Officeyutong: A successful call to `cu_kernel_get_name` implies that `name` won't be nullptr, I think

Officeyutong:
I suggest writing a short description about what `CUDA Graph` does and how this PR was designed, so we can review it more clearly @Sy0307

attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp
Sy0307: To check the segfault of old implementations. Removed.

attach/nv_attach_impl/nv_attach_impl.hpp
Sy0307: In the CUDA Driver API's Graph programming model, node creation and parameter updates (such as `cuGraphAddKernelNode` and the `SetParams` series) rely on resolved kernel handles rather than symbol names. Consequently, we must first map the CUfunction handle back to its name in order to query patched_kernel_by_name. However, repeatedly invoking the Driver API to perform this 'handle-to-name' resolution would incur significant overhead.

attach/nv_attach_impl/pass/ptxpass_core/src/core.cpp
Sy0307: OK, this is just for reviewing.

attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp
Sy0307: V2 series can. Modified.

PR #532, opened 2025-12-22
support cudaMemcpyFromSymbol

Forsworns:
## Description

The `cudaMemcpyFromSymbol` and `cudaMemcpyFromSymbolAsync` were not supported in the bpftime. The value in GPU symbols could not be retrieved to the host side.

Because we replaced the kernels, the original API calls in CUDA would read wrong device addresses. In this commit, I mimic the implementation for `cudaMemcpyToSymbol`, and replace the original CUDA API calls.

## Type of change

## How Has This Been Tested?

By adding an additional counter to `example/gpu/cuda-counter/vec_add.cu`, it can be verified that both copying from and copying to the symbol works correctly now, which increases when the kernel function is called:

image

Without this patch, the counter will always be a constant value.

**Test Configuration**:
- Hardware: L20
- Toolchain: clang 18.1
- SDK: CUDA 12.8


Officeyutong (approved):
Thanks for your contribution!

PR #560, opened 2026-03-18
feat: skip compile/load of unmodified PTX to reduce CUDA attach latency

yunwei37:
## Summary

This change reduces CUDA attach latency by only compiling and loading PTX files that were actually modified by the pass pipeline. Previously, every extracted PTX file went through compile and module-load even when the pass pipeline left it unchanged.

The implementation now propagates a per-PTX modified flag out of the patch pipeline, filters compile/load work down to modified entries, adds timing instrumentation for extract/patch/compile/load stages, and enables the required `llvm-vm` support when CUDA attach is enabled.

## Benchmarks

| Metric | Baseline | Optimized | Delta |
| --- | ---: | ---: | ---: |
| First cold fatbin | 27.9s | 5.6s | -80.0% |
| PTX compile | 21.2s | 0.17s | -99.2% |
| Fatbins 3..48 mean | 210ms | 123ms | -41.5% |

Tested on `llama.cpp` 1B with RTX 5090 on CUDA 12.9.

## Review Fixes Applied

- Added thread-safe access around `ptx_pool` and `module_pool`, plus per-fatbin `load_mutex` serialization to avoid duplicate compile/load races.
- Kept `ptx_loaded` handling aligned with the filtered load path so modified PTX modules are still marked as loaded exactly once per fatbin/device flow.

## Notes

- Probe hit behavior is a pre-existing issue and is not introduced by this optimization.
- Issue: #552

### DIFF
--- a/attach/nv_attach_impl/README-passes.md
+++ b/attach/nv_attach_impl/README-passes.md
@@ -41,22 +41,17 @@ bash example/gpu/cuda-counter/run_nv_ptx_passes.sh
   - To drive the pass order and subset via JSON configuration, set the `BPFTIME_PTXPASS_DIR` environment variable to a directory containing `*.json` configuration files. The log should print "Discovered <N> pass definitions from <dir>".
   - If the log shows "Discovered 0 pass definitions...", the system will use a fallback process (which can still complete the injection and repackaging). Check the run logs to confirm that `hack_fatbin` has been executed.
 
-### Pass Host ABI And Metadata JSON
+### Pass Standalone Executable & Configuration (JSON-only I/O)
 - **Path:** `attach/nv_attach_impl/pass/ptxpass_*`
-  - The shared-library entry point is:
-
-    ```cpp
-    int process_input(const char *ptx_text, size_t ptx_len,
-                      const char *meta_json, int meta_len,
-                      char *output, int output_len);
-    ```
-
-  - `ptx_text`/`ptx_len` carries the raw PTX blob directly and is not JSON-encoded.
-  - `meta_json` contains only small request metadata:
-    - `to_patch_kernel`
-    - `global_ebpf_map_info_symbol` (defaults to `map_info`)
-    - `ebpf_communication_data_symbol` (defaults to `constData`)
-    - `ebpf_instructions`
-  - The JSON response always includes `modified`; `output_ptx` is included only when the pass actually rewrites PTX.
-  - Legacy callers that still embed `input.full_ptx` in the metadata JSON can still be parsed by the shared core helpers, but the framework no longer emits that format.
-- **Default JSON configurations** are located at: `attach/nv_attach_impl/configs/ptxpass/*.json` (these can be copied to a custom directory and specified via `BPFTIME_PTXPASS_DIR`). 
+  - All passes communicate via JSON-structured data through stdin/stdout (plain text PTX is no longer accepted):
+    - **Input (stdin):**
+      - `full_ptx`: string, the complete PTX to be processed.
+      - `to_patch_kernel`: string, the name of the target kernel (optional).
+      - `global_ebpf_map_info_symbol`: string, defaults to `map_info`.
+      - `ebpf_communication_data_symbol`: string, defaults to `constData`.
+      - Other pass-specific fields (e.g., `source_symbol`, `copy_bytes`, `align_bytes` for memcapture).
+    - **Output (stdout):**
+      - `output_ptx`: string, the transformed PTX (an empty string or omitted field indicates no modifications).
+  - The `PTX_ATTACH_POINT` environment variable is passed to specify the current attachment point being processed.
+  - **Note:** Stdin only accepts JSON. Non-JSON input will cause the program to error out and exit immediately.
+- **Default JSON configurations** are located at: `attach/nv_attach_impl/configs/ptxpass/*.json` (these can be copied to a custom directory and specified via `BPFTIME_PTXPASS_DIR`).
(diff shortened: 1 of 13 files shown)


yunwei37:
Reviewed `25c3c10..1b60bc3` against the CUDA attach latency checklist.

No blocking correctness issues found.

- `modified_ptx` filtering in `try_loading_ptxs_for_device()` looks correct for multiple probes targeting different kernels in the same PTX/fatbin. `hack_fatbin()` produces one final PTX per input file and marks that file modified if any pass changed it, so the whole PTX stays in the compile/load set.
- The framework-level `ptx_may_contain_target_kernel()` pre-filter in `hack_fatbin()` is safe with `current_ptx` for the passes in this PR. Each pass consumes the PTX emitted by the previous pass, and the current transforms preserve untouched kernel definitions, so later kernel checks still see the remaining targets. The inline comment also explains why this logic is duplicated at the framework layer: to avoid JSON serialization and request-hash overhead for unrelated PTX blobs.
- The pass-level early return does serialize the original PTX back via `emit_runtime_response_and_return(full_ptx, false)`. `snprintf(output, length, ...)` is bounded, so this does not create a buffer overflow. The remaining risk is truncation if `length` is too small, but that is pre-existing in the normal success path as well and is not introduced by the new fast-path.
- Locking looks consistent. `load_mutex` is taken first at fatbin scope, and the new pool mutexes are only taken inside that flow; I did not find an inverse acquisition order that would deadlock. The same fatbin loading on multiple devices is therefore serialized per-fatbin rather than raced.
- `ptx_loaded = true` is still correct when every PTX is unmodified: it now means the load/patch decision for that fatbin/device completed, not that patched modules necessarily exist. The registration path still falls back to the original CUDA registration when no patched module contains the symbol.
- Timing logs are consistent enough to grep by the `GPU attach timing:` prefix. A parser can extract them today with phase-specific regexes. If we later want a single generic machine parser, switching these to a key=value schema would make that more robust, but I would treat that as non-blocking.
- I did not find formatting-only churn or unnecessary refactors in the CUDA attach code. The `vm/compat/CMakeLists.txt` change looks justified because CUDA attach links against the LLVM-based VM pieces even when `BPFTIME_LLVM_JIT` is otherwise disabled.
- Edge cases checked: empty `kernel` is handled by returning `true`; empty `hook_entries` leaves all PTX unmodified and skips compile/load; same-fatbin multi-device loads are serialized by `load_mutex`.

Local verification:
- Built `bpftime_nv_attach_tests`
- Ran `./build/attach/nv_attach_impl/test/bpftime_nv_attach_tests` successfully (`102 assertions in 24 test cases`)

No code changes required from this review.

yunwei37:
Reviewed the full \ diff against the minimal-necessary-change criteria.

Reverted:
- \: documentation-only update, not required for the optimization or new ABI to work.
- \: moved \ construction back inside the kernel loop; the hoist was a small micro-optimization/churn item, not required for correctness or the main latency win.

Kept:
- \, \, the 3 pass entrypoints, and the unit-test updates: required for the pass ABI split (raw PTX + metadata JSON), modified/unmodified response handling, correct cache-keying, and skipping compile/load for unmodified PTX.
- \: required for skip-compile/load on unmodified PTX, attach timing instrumentation, and mutex coverage for \, \, and per-fatbin load serialization.
- \: kept the timing logs; kept WARN->DEBUG demotion because the new expected "symbol not present in a patched module" path would otherwise spam warnings when original CUDA registration is intentionally retained.
- \: required when \ and \; CUDA attach still depends on \, so this is not cosmetic.
- \: kept because the pass-internal fast path is still active in all 3 passes. The helper is small and avoids duplicating the same pre-check in each pass.

No other formatting-only or cosmetic churn stood out as worth reverting after the above two changes.

Resulting diff vs base is now 12 files, +503/-154.

Validation:
- \: passed.
- \: passed (108 assertions in 24 test cases).
- \ run: the optimization path behaved as expected. First large fatbin showed extract 521 ms, patch 104 ms for 120 PTX files with 1 modified, compile 166 ms for 1 PTX, module load 1 ms, total attach 272 ms. Repeated later fatbins showed the intended fast path: \ for unmodified PTX and \.
- The exact \ command exited 124. Target log reached \ and the loader-side \ counters stayed 0 throughout the run, so functional verification is inconclusive with that timeout and did not demonstrate the probe firing.

yunwei37:
Reviewed the full `25c3c10..HEAD` diff against the minimal-necessary-change criteria.

Reverted:
- `attach/nv_attach_impl/README-passes.md`: documentation-only update, not required for the optimization or new ABI to work.
- `attach/nv_attach_impl/nv_attach_impl.cpp`: moved `ebpf_inst_words` construction back inside the kernel loop; the hoist was a small micro-optimization/churn item, not required for correctness or the main latency win.

Kept:
- `attach/nv_attach_impl/nv_attach_impl.cpp/.hpp`, `ptxpass_core`, the 3 pass entrypoints, and the unit-test updates: required for the pass ABI split (raw PTX + metadata JSON), modified/unmodified response handling, correct cache-keying, and skipping compile/load for unmodified PTX.
- `attach/nv_attach_impl/nv_attach_fatbin_record.cpp/.hpp`: required for skip-compile/load on unmodified PTX, attach timing instrumentation, and mutex coverage for `ptx_pool`, `module_pool`, and per-fatbin load serialization.
- `attach/nv_attach_impl/nv_attach_impl_frida_setup.cpp`: kept the timing logs; kept WARN->DEBUG demotion because the new expected "symbol not present in a patched module" path would otherwise spam warnings when original CUDA registration is intentionally retained.
- `vm/compat/CMakeLists.txt`: required when `BPFTIME_ENABLE_CUDA_ATTACH=1` and `BPFTIME_LLVM_JIT=0`; CUDA attach still depends on `llvmbpf_vm`, so this is not cosmetic.
- `ptx_may_contain_target_kernel()`: kept because the pass-internal fast path is still active in all 3 passes. The helper is small and avoids duplicating the same pre-check in each pass.

No other formatting-only or cosmetic churn stood out as worth reverting after the above two changes.

Resulting diff vs base is now 12 files, +503/-154.

Validation:
- `cmake --build build -j$(nproc)`: passed.
- `./build/attach/nv_attach_impl/test/bpftime_nv_attach_tests`: passed (108 assertions in 24 test cases).
- `llama.cpp` run: the optimization path behaved as expected. First large fatbin showed extract 521 ms, patch 104 ms for 120 PTX files with 1 modified, compile 166 ms for 1 PTX, module load 1 ms, total attach 272 ms. Repeated later fatbins showed the intended fast path: `Skipping PTX compile/load` for unmodified PTX and `PTX compile took 0-2 ms`.
- The exact `timeout 120s` command exited 124. Target log reached `ggml_cuda_init` and the loader-side `threadhist` counters stayed 0 throughout the run, so functional verification is inconclusive with that timeout and did not demonstrate the probe firing.

PR #574, opened 2026-05-26
fix: enable bpf_cookie support by fixing three  bugs

### LINKED ISSUE #569: [BUG] bpf_link_handler constructor does not initialize attach_cookie from bpf_cookie
**Describe the bug**
here are two related bugs in the bpf_link code path:

bpf_cookie not initialized: The bpf_link_handler(bpf_link_create_args) constructor in runtime/src/handler/link_handler.hpp fails to initialize the attach_cookie member from args.perf_event.bpf_cookie. Since attach_cookie is of type std::optional<uint64_t>, it is left in a disengaged state (empty). Consequently, bpf_get_attach_cookie() always returns 0 for BPF programs attached through the add_bpf_link() code path, even when a valid bpf_cookie was supplied in the BPF_LINK_CREATE syscall.

FEAT_PERF_LINK detection fails: In runtime/src/bpftime_shm_internal.cpp, the add_bpf_link() function does not validate whether target_fd is a valid perf fd when attach_type is BPF_PERF_EVENT (41).

**To Reproduce**
Write a BPF program that invokes bpf_get_attach_cookie() and stores the result in a map
Attach the program to a uprobe with a non-zero cookie via bpf_program__attach_uprobe_opts() with .cookie set
Trigger the probe target function
Read the map value — it returns 0 instead of the configured cookie

**Expected behavior**
bpf_get_attach_cookie() should return the cookie value supplied at attach time, consistent with kernel behavior.

**Screenshots**

N/A

**Desktop (please complete the following information):**

OS: Linux（NingOS）
Version: bpftime master (commit fac864e5c0775756323794ff44e03ab79a11e753)

**Additional context**
core code：

```cpp
// Before — attach_cookie is default-constructed (disengaged):
bpf_link_handler(struct bpf_link_create_args args)
    : args(args), prog_id(args.prog_fd),
      attach_target_id(args.target_fd)
{}

// After — attach_cookie is initialized from the bpf_link_create_args:
bpf_link_handler(struct bpf_link_create_args args)
    : args(args), prog_id(args.prog_fd),
      attach_target_id(args.target_fd),
      attach_cookie(args.perf_event.bpf_cookie)
{}
```

BorgesLu:
Fixes #569

## Remaining Fix

The fallback path in `syscall_context.cpp` discards the return value of `bpftime_attach_perf_to_bpf_with_cookie()`, assigning it to neither the local variable `id` nor any other storage. This causes link fd loss and state inconsistency when the fallback path is triggered.

## Already Fixed Upstream

The other two bugs from the original PR have already been fixed upstream:
1. **FEAT_PERF_LINK detection** — `add_bpf_link()` now validates `target_fd` for `BPF_PERF_EVENT` (fixed in #xxx)
2. **attach_cookie initialization** — `bpf_link_handler` constructor now initializes `attach_cookie` from `args.perf_event.bpf_cookie` (fixed in #xxx)

## Type of change

### DIFF
--- a/runtime/syscall-server/syscall_context.cpp
+++ b/runtime/syscall-server/syscall_context.cpp
@@ -594,8 +594,7 @@ long syscall_context::handle_sysbpf(int cmd, union bpf_attr *attr, size_t size)
 			SPDLOG_DEBUG(
 				"Attaching perf event {} to prog {}, with bpf cookie {:x}",
 				target_fd, prog_fd, cookie);
-			bpftime_attach_perf_to_bpf_with_cookie(target_fd,
-							       prog_fd, cookie);
+				id = bpftime_attach_perf_to_bpf_with_cookie(target_fd, prog_fd, cookie);
 		}
 		return id;
 	}


maxgio92:
It seems that this PR needs to be rebased on main, now that #570 landed.

I guess that the remaining parts that still can be useful are the `id` returned by `bpftime_attach_perf_to_bpf_with_cookie` and the new test suite for the attach with cookie path

BorgesLu:
@maxgio92  Thanks for the review! I've rebased on main and narrowed this PR to only the `id` return value fix, since the other two bugs are already fixed by #570.

yunwei37:
Thanks a lot @BorgesLu @maxgio92 !

PR #575, opened 2026-05-26
runtime: shard software perf event producers by thread

yunwei37:
## Summary

Move userspace software perf event output away from a single per-CPU producer buffer and into per-thread producer shards stored in shared memory. Each producer thread writes only its own SPSC shard, while the existing consumer-facing mmap buffer remains the public read surface and is populated from shards during readiness checks.

This keeps the hot producer path lock-free after first-thread shard registration and avoids the current race where multiple userspace threads pinned to the same CPU can observe the same `data_head` and overwrite each other.

## Details

- Adds `software_perf_event_buffer` as the reusable perf ring implementation.
- Adds shared-memory per-thread `software_perf_event_shard` storage under each software perf event.
- Keeps `bpftime_get_software_perf_event_raw_buffer()` compatible by returning the existing logical consumer mmap buffer.
- Drains producer shards into the consumer buffer from `has_data()`, which is already used by the mocked epoll readiness path.
- Adds a concurrency unit test with multiple threads writing the same logical software perf event and verifies each record appears exactly once.

## Validation

- `cmake --build build --target runtime -j$(nproc)`
- `cmake -S . -B build-perf-test -DBPFTIME_ENABLE_UNIT_TESTING=ON -DBPFTIME_ENABLE_CUDA_ATTACH=OFF -DCMAKE_BUILD_TYPE=RelWithDebInfo`
- `cmake --build build-perf-test --target bpftime_runtime_tests -j$(nproc)`
- `./build-perf-test/runtime/unit-test/bpftime_runtime_tests "[perf_event][software_perf_event]"`
- `./build-perf-test/runtime/unit-test/bpftime_runtime_tests '~Test tail calling from userspace to kernel' '~Test shm progs attach'`

Full `bpftime_runtime_tests` still has two environment-sensitive failures in this checkout: `Test shm progs attach` fails on a file-equivalence check with an empty path, and `Test tail calling from userspace to kernel` fails because map creation returns `-1` with errno `1`. The remaining 69 runtime test cases pass.

### DIFF
--- a/runtime/src/handler/perf_event_handler.cpp
+++ b/runtime/src/handler/perf_event_handler.cpp
@@ -17,7 +17,13 @@
 #include <boost/interprocess/allocators/allocator.hpp>
 #include <unistd.h>
 #include <spdlog/fmt/bin_to_hex.h>
+#include <unordered_map>
 #include <variant>
+#if __linux__
+#include <sys/syscall.h>
+#elif __APPLE__
+#include <pthread.h>
+#endif
 
 #pragma GCC diagnostic ignored "-Wstrict-aliasing"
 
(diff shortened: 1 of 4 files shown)


Officeyutong:
Is this implementation consistent with kernel's? Or does it need to be consistent with kernel's implementation?

yunwei37:
No. In userspace, the schedule unit is different than kernel, so I think the thread local maps would be the best choice.

If you don't do this, you can pin 2 thread to one CPU to cause bugs.

It can still keep compatible with kernel per cpu maps, but the semantics are different

PR #597, opened 2026-07-14
fix(runtime): handle shared memory exhaustion

### LINKED ISSUE #596: [BUG] Unhandled exceptions for insufficient BPFTIME_SHM_MEMORY_MB
**Description of the bug**
I observed crash in two places when sufficient BPFTIME_SHM_MEMORY_MB size/value is not passed.
1) Creation of handler_variant_vector handlers; in handler _manager causes crash.
2) Perf buffer creation causes crash when insufficient BPFTIME_SHM_MEMORY_MB is set for higher Perf_buffer_pages.

**To Reproduce**
Case 1:
This can be reproduced with malloc example mentioned in this repo (no changes to malloc example).

Steps:
 - set DEFAULT_MAX_FD_COUNT in bpftime_config.hpp to 1024 * 1024 (re-compile)
 - export BPFTIME_SHM_MEMORY_MB=64 before starting execution.

Case 2:
This can be reproduced with bashreadline example mentioned in this repo (PERF_BUFFER_PAGES  should  set to appropriate value based on machine configuration)
Assumptions (Values to be modified based on the machine configuration):
 - Page_size: 4096
 - Perf_buffer_pages: 4096
 - num_CPUs: 256
Min BPFTIME_SHM_MEMORY_MB required only for perf event buffers: 4GB
(Page_size x Perf_buffer_pages x num_CPUs)

Steps:
 - set appropriately large value for page count in perf_buffer__new()
 - export BPFTIME_SHM_MEMORY_MB=3072

**Expected behavior**
Case1: Exit with code 1 instead of abort().
Case 2: perf_buffer__new() call should return NULL.

**Desktop :**

* OS: [Ubuntu]
* Version [24.04.1 LTS (Noble Numbat)]

yunwei37:
## Summary

- exit cleanly with status 1 and an actionable message when syscall-server startup cannot allocate its configured shared-memory state
- return `MAP_FAILED` with deterministic `ENOMEM` when a software perf-event buffer cannot grow, without changing allocation handling for unrelated syscall paths
- validate perf ring sizes before mutation and document the relationship between shared-memory size and fd-table capacity
- add real `LD_PRELOAD` process tests for both startup failure and the exported `mmap()` boundary

## Testing

- `cmake --build build --target bpftime_runtime_tests -j2`
- `./build/runtime/unit-test/bpftime_runtime_tests "[allocation]"` (2 test cases, 6 assertions)
- `./build/runtime/unit-test/bpftime_runtime_tests "[perf_event]"` (3 test cases, 8279 assertions)
- unfiltered local suite: 76 test cases, 74 passed; the same two environment-dependent failures occur on the unchanged baseline (`Test shm progs attach` and `Test tail calling from userspace to kernel`)
- `git diff --check`

Closes #596

### DIFF
--- a/runtime/src/handler/perf_event_handler.cpp
+++ b/runtime/src/handler/perf_event_handler.cpp
@@ -9,7 +9,6 @@
 #include "bpftime_epoll.h"
 #endif
 #include "spdlog/spdlog.h"
-#include <boost/container/throw_exception.hpp>
 #include <boost/interprocess/detail/segment_manager_helper.hpp>
 #include <boost/interprocess/smart_ptr/shared_ptr.hpp>
 #include <cstring>
@@ -284,9 +283,7 @@ void *software_perf_event_buffer::ensure_mmap_buffer(size_t buffer_size)
 		};
 		try {
 			mmap_buffer.resize(buffer_size);
-		} catch (const boost::interprocess::bad_alloc &e) {
-			return report_allocation_failure(e);
-		} catch (const boost::container::length_error_t &e) {
+		} catch (const std::exception &e) {
 			return report_allocation_failure(e);
 		}
 		// Update data size in the mmap header
--- a/runtime/src/handler/perf_event_handler.cpp
+++ b/runtime/src/handler/perf_event_handler.cpp
@@ -9,6 +9,7 @@
 #include "bpftime_epoll.h"
 #endif
 #include "spdlog/spdlog.h"
+#include <boost/container/throw_exception.hpp>
 #include <boost/interprocess/detail/segment_manager_helper.hpp>
 #include <boost/interprocess/smart_ptr/shared_ptr.hpp>
 #include <cstring>
@@ -19,6 +20,7 @@
 #include <spdlog/fmt/bin_to_hex.h>
 #include <atomic>
 #include <cerrno>
+#include <exception>
 #include <limits>
 #include <unordered_map>
 #include <variant>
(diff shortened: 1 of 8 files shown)


yunwei37:
### Independent review result

Reviewed final head `6caa6d595cbefc27dbdbd78d984d62858da8500c` after two review/fix rounds.

Initial blockers were: overly broad recoverable allocation handling, missing process-boundary coverage, a startup test that could pass through the pre-existing outer exception wrapper, and stale `errno` propagation. The final head reverts the broad syscall behavior change, drives startup through the unwrapped `open()` entrypoint, exercises exported `mmap()` with a stale incoming errno, and verifies clean `exit(1)` / `MAP_FAILED` + `ENOMEM`. The scoped resize catch is compatible across Boost versions.

Final verdict: **no findings and no blockers**. `git diff --check`, the build, and focused allocation tests pass locally.

yunwei37:
### Claude cross-agent final-head review

Reviewed the immutable patch for final head `6caa6d595cbefc27dbdbd78d984d62858da8500c` against base `6bd5319d3afb683716c7f9273508598079a8dc7e`.

**Verdict: NO BLOCKERS.**

The review verified that the prior-round fixes are present and effective:

- startup allocation failure is caught at the unwrapped interposer boundary and exits cleanly;
- the catch uses the public Boost.Interprocess `bad_alloc` type;
- the process helper genuinely exercises unwrapped `open()`;
- stale incoming `errno` cannot make the mmap regression test pass spuriously;
- size validation happens before mutation, and the resize catch is scoped to the allocation operation;
- the earlier broad syscall exception behavior change is absent.

One medium-severity hardening suggestion was recorded as non-blocking: the mmap path relays `errno` through the runtime boundary rather than using an explicit error out-parameter. Low-severity notes concern pre-existing non-allocation startup behavior, future test assumptions, and harmless defensive code.

Follow-up verification resolved the review's two doc/log checks: `DEFAULT_MAX_FD_COUNT` is `1024 * 6` (6144), and `shm_memory_size` is explicitly configured in MB, consistent with the startup diagnostic.

No code changes are required from this review.

yunwei37:
CI triage update (2026-07-16): the only current blocking checks are `build-bpftime-gpu` and `test-ptxpass`, both queued on `[self-hosted, Linux, X64, gpu]`. This matches the same runner-unavailable pattern as #598/#586/#601, not a code failure visible in logs. Needs GPU self-hosted runner availability before CI can go green.

yunwei37:
CI triage update:

- The only current blockers I can reproduce from the PR checks are build-bpftime-gpu and test-ptxpass.
- Both jobs are queued on the self-hosted [Linux, X64, gpu] runner and match the same long queued/cancelled pattern seen across other PRs.
- I do not see evidence of a branch-specific code failure in the available logs.

Action needed: restore or free the GPU self-hosted runner, then rerun these two workflows. No code change pushed to this PR.

Issue #449, opened 2025-09-17
[Need Help] bpftime-daemon can't run


Yifei-Zhang-Deeproute:
I was able to compile successfully in my container environment, but I can't run bpftime_daemon.

The error problems encountered are:
```
[2025-09-17 08:19:46.825] [info] [bpf_tracer.cpp:297] /a 已存在
[2025-09-17 08:19:46.827] [info] [bpftime_shm_internal.cpp:714] Global shm constructed. shm_open_type 0 for bpftime_maps_shm
[2025-09-17 08:19:46.827] [info] [bpftime_shm_internal.cpp:40] Global shm initialized
libbpf: prog 'tracepoint__syscalls__sys_enter_bpf': relo #12: target candidate search failed for [30] struct : -22
libbpf: prog 'tracepoint__syscalls__sys_enter_bpf': relo #12: failed to relocate: -22
libbpf: failed to perform CO-RE relocations: -22
libbpf: failed to load object 'bpf_tracer_bpf'
libbpf: failed to load BPF skeleton 'bpf_tracer_bpf': -22
failed to load BPF object: -22
```

After my line-by-line comment troubleshooting, I found that the problem was caused by should_modify_program function（bpf_tracer.bpf.c）:
```
static bool should_modify_program(union bpf_attr *new_attr)
{
	if (new_attr->insn_cnt < 2 || !enable_replace_prog) {
		return false;
	}

	if (new_attr->prog_type != BPF_PROG_TYPE_KPROBE &&
	    new_attr->prog_type != BPF_PROG_TYPE_TRACEPOINT) {
		return false;
	}

	return true;
}
```
But I'm not clear why there is a CO-RE relocations issue here.
Can you give some advice on how to solve this problem? And why is this happening? Thank You！！

YRXING:
> I was able to compile successfully in my container environment, but I can't run bpftime_daemon.
> 
> The error problems encountered are:
> 
> ```
> [2025-09-17 08:19:46.825] [info] [bpf_tracer.cpp:297] /a 已存在
> [2025-09-17 08:19:46.827] [info] [bpftime_shm_internal.cpp:714] Global shm constructed. shm_open_type 0 for bpftime_maps_shm
> [2025-09-17 08:19:46.827] [info] [bpftime_shm_internal.cpp:40] Global shm initialized
> libbpf: prog 'tracepoint__syscalls__sys_enter_bpf': relo #12: target candidate search failed for [30] struct : -22
[...]

PR #602, opened 2026-07-19
fix: avoid stack-copy CO-RE field reads

yunwei37:
## Summary

- read `prog_type` from the original user `bpf_attr` pointer with `BPF_CORE_READ_USER`
- pass `insn_cnt` and `prog_type` scalars into `should_modify_program`
- avoid direct CO-RE field reads on the stack copy while preserving the existing program replacement path

## Validation

- `clang-format --dry-run --Werror daemon/kernel/bpf_tracer.bpf.c`
- `git diff --check`
- configured with warnings as errors and built `bpftime_daemon_ebpf_target` successfully

Refs #449

### DIFF
--- a/daemon/kernel/bpf_tracer.bpf.c
+++ b/daemon/kernel/bpf_tracer.bpf.c
@@ -154,7 +154,8 @@ static int process_bpf_prog_load_events(union bpf_attr *attr)
 	unsigned int prog_type;
 	void *insns;
 	union bpf_attr new_attr = { 0 };
-	bpf_probe_read(&new_attr, sizeof(new_attr), attr);
+	long attr_read_result =
+		bpf_probe_read_user(&new_attr, sizeof(new_attr), attr);
 
 	insn_cnt = BPF_CORE_READ_USER(attr, insn_cnt);
 	prog_type = BPF_CORE_READ_USER(attr, prog_type);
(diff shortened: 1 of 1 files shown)


yunwei37:
Patrol review and validation summary for exact head `2948eb4`:

- The original issue was reproduced at the object-relocation level: three predicate reads were being emitted as CO-RE accesses through the stack-copied `union bpf_attr`. The patch now reads `insn_cnt` and `prog_type` from the original userspace pointer and passes scalar values into the predicate, removing those relocations.
- Dedicated review found one P2 edge case in the first commit: replacement could proceed when the full userspace attr snapshot failed. Follow-up `2948eb4` switches that copy to `bpf_probe_read_user` and requires a successful full snapshot before either write. Exact-head re-review and disassembly confirm the finding is closed.
- A separate different-family read-only review of the final two-commit range also returned APPROVE; it independently confirmed the scalar CO-RE behavior, the corrected snapshot gate, and no new verifier/stack blocker.
- Local validation passes: formatting/diff checks, warnings-as-errors CMake configuration, and `bpftime_daemon_ebpf_target` build. Exact-head GCC 13, Docker image build, CodeQL, and secret scanning are green. Maximum BPF stack is 352 bytes, below the 512-byte limit.

The remaining coverage gap is a non-blocking load smoke on the reporter's exact BTF/compiler combination (and a synthetic short-attr case). All current-head GitHub checks pass; maintainer review is still required.

PR #613, opened 2026-07-30
fix: close CUDA CLI command branch

yunwei37:
## Summary

- close the CUDA-only `run-on-cuda` command branch before the fallback `else`
- restore compilation of `bpftimetool` when `BPFTIME_ENABLE_CUDA_ATTACH=ON`

## Validation

- `cmake -B /tmp/bpftime-cuda-cli.JlWUeK -S . -DCMAKE_BUILD_TYPE=Debug -DBPFTIME_LLVM_JIT=YES -DBPFTIME_ENABLE_CUDA_ATTACH=ON -DBPFTIME_CUDA_ROOT=/usr/local/cuda -DLLVM_DIR=/usr/lib/llvm-15/lib/cmake/llvm`
- `cmake --build /tmp/bpftime-cuda-cli.JlWUeK --target bpftimetool -j2`

The missing brace was exposed by the GPU build job in PR #612.

### DIFF
--- a/tools/bpftimetool/main.cpp
+++ b/tools/bpftimetool/main.cpp
@@ -371,6 +371,7 @@ int main(int argc, char *argv[])
 			}
 			return 1;
 		}
+	}
 	#endif
 	else {
 		cerr << "Invalid subcommand " << cmd << endl;


yunwei37:
Reviewed exact head `64cbd480edad93a3edf7594bd089e6ee28c062e0` with both required independent passes. Neither found a blocker. Both verified that the added brace closes only the CUDA-gated `run-on-cuda` branch; the CUDA-disabled preprocessed control flow is unchanged.

The diff is one production line (`+1/-0`) with no helper, fixture, test, workflow, API, default, dependency, or documentation change, so it is already the minimum possible fix. A local CUDA-enabled `bpftimetool` build passes, and the previously failing `build-bpftime-gpu` check is now green. Remaining checks are still running. GitHub reviews, inline comments, review threads, and Copilot comments were also checked; none are present.

yunwei37:
为啥不放在那个 branch? 干嘛不关掉这个?

PR #614, opened 2026-07-30
fix: avoid duplicate LLVM VM linkage in CUDA agent

yunwei37:
## Summary

- stop whole-archiving `bpftime_llvm_vm` a second time in the CUDA agent link block
- retain the directly embedded LLVM factory object and the existing normal target dependency

This is stacked on #604 because that pull request removes the same duplicate linkage from the common agent path. Relative to #604, this changes one CMake link entry.

## Validation

- configured with `BPFTIME_ENABLE_CUDA_ATTACH=ON` and CUDA 12.9
- built `bpftime-agent` successfully
- inspected the generated link command: the LLVM factory object remains directly embedded, while `libbpftime_llvm_vm.a` is no longer inside the CUDA whole-archive group

### DIFF
--- a/runtime/agent/CMakeLists.txt
+++ b/runtime/agent/CMakeLists.txt
@@ -36,7 +36,7 @@ endif()
 
 if(${BPFTIME_ENABLE_CUDA_ATTACH})
   add_dependencies(bpftime-agent bpftime_llvm_vm llvmbpf_vm bpftime_vm)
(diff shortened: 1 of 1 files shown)


yunwei37:
The exact `8a4948d` head has now been checked in both the default uBPF-enabled CUDA configuration and the uBPF-disabled matrix that exposed the duplicate factory definitions. With CUDA 12.9 and LLVM enabled, `bpftime-agent` links successfully; `create_llvm_vm_instance()` and `register_llvm_vm_factory()` are each retained once.

Two independent read-only reviews found no blockers. The diff remains one CMake line: it removes `libbpftime_llvm_vm.a` only from the CUDA whole-archive group while preserving the directly embedded factory object, ordinary target dependencies, and the other CUDA archives. There are no review threads or Copilot comments.

This PR is intentionally stacked on #604, so the remaining actions are the base PR and maintainer review.

PR #586, opened 2026-06-21
daemon: preserve mirrored perf-link cookies

yunwei37:
## Summary

- preserve `perf_event.bpf_cookie` when the daemon mirrors a `BPF_PERF_EVENT` link
- pass the optional cookie through `bpftime_driver` to the existing cookie-aware runtime attach API
- replace the placeholder daemon-driver test with a regression that inspects the mirrored link handler cookie

## Scope

This PR is intentionally limited to the daemon propagation gap. The earlier OTel example, benchmark, map types, CPU pinning, register changes, and lifecycle fixes were removed because they are separate concerns and the example only validated native collector output, not bpftime execution.

## Validation

- `cmake --build build-pr586 --target bpftime_daemon_tests -j$(nproc)`
- `build-pr586/daemon/test/bpftime_daemon_tests` (2 test cases, 4 assertions passed)
- `git clang-format --diff origin/master` (no changes)
- `git diff --check origin/master HEAD`

Final net diff: 4 files, `+57/-26`, reduced from 31 files, `+1231/-84`.

Fixes #616

PR #623, opened 2026-08-01
example: add OTel profiler mirror compatibility smoke test

yunwei37:
## Summary

- add a runnable OpenTelemetry eBPF profiler compatibility example for libc `malloc`/`free`
- verify that the bpftime daemon mirrors both cilium/ebpf perf-event `BPF_LINK_CREATE` uprobes and successfully attaches them
- pin the tested upstream profiler revision and export a real profile containing the example victim

## Scope and minimality

This is intentionally a daemon-mirror compatibility example and a follow-up to #586. The upstream Go collector loads and executes its profiler programs in the kernel, so the exported profiles are native-kernel output. The example does not claim that the full upstream profiler runs through the bpftime agent; its map and helper requirements are not all supported yet.

The final change is example-only: 6 files, `+188/-0`, with no runtime, daemon, VM, agent, attach, API, dependency, or CI-matrix changes.

Reduction trail:

- initial: 7 files, +204
- removed a redundant single-rule Makefile: -13
- removed the busy victim's unused integer state/include: -3
- final: 6 files, +188

The existing sleeping malloc victim was tested with the same pinned collector: bpftime observed both handlers and attaches, but the collector emitted zero victim profiles in 12 seconds. The 13-line busy victim is therefore retained as the smallest reliable sampling workload.

## Validation

Exact head: `36fe67e01dafb3ed2a532ebbda84f9559a7bfe43`

- built upstream `open-telemetry/opentelemetry-ebpf-profiler` at `39d09634147c707f035ddda78186054bfe037629` with `make ebpf otelcol-ebpf-profiler`
- built `bpftime_daemon` and the local victim
- four complete root E2E runs with the pinned collector; the final trimmed victim produced:
  - 4 `BPF_LINK_CREATE` events
  - 2 exact libc uprobe handlers
  - 2 successful bpftime attaches
  - 5 exported `victim` profile records
  - collector and daemon alive through final assertions
- `bash -n`
- C `-O2 -Wall -Wextra` compile/syntax checks
- GCC and Clang assembly both retain `malloc`/`free` with the constant volatile store
- `clang-format --dry-run --Werror`
- YAML structure parse
- `git diff --check`
- exact-head dedicated subagent review: PASS
- exact-head independent external read-only review: PASS
- exact-head Copilot comments/threads audit: no findings

## CI / draft status

The final-head CI reached terminal state. Both no-JIT runtime builds fail on the same untouched master error fixed by one-line PR #626: `runtime/syscall-server/syscall_context.cpp` still calls removed `construct_agent_config_from_env`; the two JIT matrix siblings were then canceled by fail-fast. CodeQL and GitGuardian are green. This PR stays draft until #626 merges, then it will be synchronized and the complete exact-head review/Copilot/terminal-green CI gate will run again.


yunwei37:
Exact-head review for `e436d8d0b056d57b759a9bb47a2f0a3524a328de`:

- dedicated full-diff subagent review: **PASS**
- independent detached-clone review with no remote: **PASS**
- full scope: 7 files, `+204/-0`, example-only
- pinned upstream profiler build and real daemon-mirror smoke: **PASS**
- verified two exact link events, two handlers, two successful attach events, final process liveness, and a `victim` profile
- `bash -n`, `shellcheck`, C formatting, and `git diff --check`: **PASS**

No blocking findings remain. The PR stays draft while current-head CI and GitHub review complete.

yunwei37:
@copilot review

yunwei37:
Current-head CI failure is a pre-existing master regression, not an example change: `runtime/syscall-server/syscall_context.cpp` still calls the renamed `construct_agent_config_from_env()`. The same failure is fixed by the one-line current-head patch in #604. This Draft will stay blocked rather than duplicate that unrelated runtime fix; after #604 lands, I will sync this branch and rerun the full matrix.

yunwei37:
Dependency update: the one-line master regression is now isolated in #626 (`construct_agent_config_from_env()` → `construct_runtime_config_from_env()`). #604 is the older no-libbpf PR and is no longer the prerequisite for this example. The exact-head OTel review remains valid because #623 has not changed; after #626 merges I will sync/rerun current-head CI and repeat any review required by a new SHA.

Issue #605, opened 2026-07-22
LD_PRELOAD shims must remain transparent on internal failures


yunwei37:
The preload libraries run inside an unrelated host process, so internal bpftime failures must not terminate that process or write diagnostics to its standard streams unless console logging was explicitly selected.

Current violations include:
- shared-memory exhaustion paths call exit and terminate the host;
- agent exception and CUDA teardown paths write directly to stderr;
- early logging can reach the default stderr logger before the configured file sink is installed;
- logger setup is noexcept but can throw, which can terminate the host;
- the syscall transformer exits on recoverable setup failures.

Acceptance criteria:
- preload initialization and interception failures fail open to the original host operation where possible;
- no exception crosses an exported C/interposition boundary;
- no preload path calls exit or abort because of an internal bpftime error;
- stdout and stderr remain unchanged by default; console output requires explicit BPFTIME_LOG_OUTPUT=console;
- logger sink failures fall back to a silent sink instead of stderr or process termination;
- regression tests cover shared-memory exhaustion, host exit status, and default stdio silence;
- the invariant is recorded in repository maintenance guidance and a dedicated design document without changing the main README.

yunwei37:
The fix is available in #606. It records the invariant in the root maintenance guidance, adds fail-open handling across the preload components, makes logger failures silent, and adds regression coverage for host exit status, stdio, errno, mmap/openat fallback, and transformer page-permission cleanup. The pull request remains open for checks and maintainer review.

AI-generated response; a maintainer will review and follow up later

yunwei37:
Update: #606 was intentionally narrowed after review to one independently mergeable syscall-callback failure path (2 files, +58/-16). It no longer claims to close this broader issue.

The remaining logger/allocation, syscall-server, agent lifecycle, transformer, and ABI items should be implemented as separate minimal PRs with focused validation rather than one mixed change. This issue remains open.

AI-generated response; a maintainer will review and follow up later

yunwei37:
The syscall-callback portion has now merged through the narrower replacement #628; #606 was closed without merge. The merged change contains exceptions within the callback loop, preserves override semantics, and verifies that later callbacks still run and thread-local override state is cleared after a throw.

This resolves only that independently testable slice and does not close the broader preload-transparency issue. Logger/allocation failure paths, syscall-server and agent lifecycle handling, transformer behavior, the maintenance-guidance requirement, and the default-stdio invariant remain separate follow-up work.

[...]

PR #632, opened 2026-08-08
fix(runtime): keep preload logging off host stdio

yunwei37:
Part of #605.

## Scope

This is the first of four small preload-transparency fixes. It only hardens common logger setup and the syscall-server early logging path. It does not remove syscall/transformer exits or assertions, and it does not change interception fallback behavior.

## Changes

- resolve original syscall functions before initializing spdlog
- keep syscall mocking disabled until logger setup, config parsing, lock initialization, and constructor logging complete
- use the configured file logger directly on the normal path; install a null fallback only if the target is invalid or setup fails
- require `BPFTIME_LOG_OUTPUT=console` before writing to host stderr
- keep invalid/unwritable paths, runtime sink errors, and disabled flushes off host stdout/stderr
- preserve global and named `SPDLOG_LEVEL` behavior on the configured sink

## Validation

- `cmake --build build --config Debug --target bpftime_runtime_tests -j4`
- `bpftime_runtime_tests "[logging]"`: 5 test cases, 32 assertions passed
- `bpftime_runtime_tests "[allocation]"`: 7 test cases, 43 assertions passed
- `git diff --check`
- local full runtime run on the initial patch: 80/82 cases passed; the two failures were the environment-dependent Frida executable-equivalence and kernel tailcall tests, outside this patch

The regression suite covers default/unwritable output, `/dev/full`, global `SPDLOG_LEVEL`, named `stderr=off`, positive explicit-console output, and complete startup logging through a valid file sink without constructor re-entry.

## Diff size

5 files, +196/-52. Production code is +67/-35 (net +32); tests are +129/-17.

### DIFF
--- a/runtime/include/bpftime_logger.hpp
+++ b/runtime/include/bpftime_logger.hpp
@@ -1,6 +1,7 @@
 #include <string>
 #include "spdlog/spdlog.h"
 #include "spdlog/cfg/env.h"
+#include "spdlog/sinks/null_sink.h"
 #include "spdlog/sinks/rotating_file_sink.h"
 #include "spdlog/sinks/stdout_color_sinks.h"
 #include <cstdlib>
@@ -38,16 +39,27 @@ inline std::string expand_user_path(const std::string &input_path)
 inline void bpftime_set_logger(const std::string &target) noexcept
 {
 	auto silence_logger = []() noexcept {
-		if (auto *logger = spdlog::default_logger_raw();
-		    logger != nullptr)
+		try {
+			auto sink =
+				std::make_shared<spdlog::sinks::null_sink_mt>();
+			auto logger = std::make_shared<spdlog::logger>(
+				"bpftime_quiet", std::move(sink));
 			logger->set_level(spdlog::level::off);
+			spdlog::set_default_logger(std::move(logger));
+		} catch (...) {
+			try {
+				if (auto *logger = spdlog::default_logger_raw();
+				    logger != nullptr)
+					logger->set_level(spdlog::level::off);
+			} catch (...) {
+			}
+		}
 	};
+	silence_logger();
 	try {
 		std::string logger_target = expand_user_path(target);
-		if (logger_target.empty()) {
-			silence_logger();
+		if (logger_target.empty())
 			return;
-		}
 
 		std::shared_ptr<spdlog::sinks::sink> sink;
 		if (logger_target == "console") {
@@ -87,7 +99,8 @@ inline void bpftime_set_logger(const std::string &target) noexcept
 inline void bpftime_logger_flush() noexcept
 {
 	try {
-		if (auto logger = spdlog::default_logger(); logger != nullptr)
+		if (auto logger = spdlog::default_logger();
+		    logger != nullptr && logger->level() != spdlog::level::off)
 			logger->flush();
 	} catch (...) {
 	}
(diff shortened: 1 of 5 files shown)


yunwei37:
Independent code review record for `6e6658c5bcd5d8ee9796be6a417de7f49dbf82c3`

**Reviewer verdict: PASS**

Verified against base `ec26daecc8e787fb80fd95dd596a576404a5e36e`:

- no spdlog use remains inside `init_original_functions()`
- mocking stays disabled across null/configured sink construction, configuration logging, spinlock initialization, and all constructor logging
- valid-file `fopen`/allocation paths call resolved originals instead of re-entering `start_up()`
- default/unwritable, `/dev/full`, global `SPDLOG_LEVEL`, named `stderr=off`, and disabled flush paths remain off host stdio
- explicit console remains opt-in and retains the `stderr` logger name
- valid-file reproduction captured zero stdout/stderr bytes and contained both startup boundary messages

Focused validation: `[logging]` 32 assertions/5 cases; `[allocation]` 43 assertions/7 cases. No avoidable scope or test-fixture growth found; no blocking findings.

yunwei37:
Independent external-model review record for `6e6658c5bcd5d8ee9796be6a417de7f49dbf82c3`

**External verdict: PASS**

Read-only review inspected the exact diff, syscall wrappers/handlers, and bundled spdlog registry/logger implementation. It confirmed:

- original libc functions resolve before the first spdlog registry/sink operation
- the constructor-wide mocking gate prevents configured-file `fopen`/mmap/open/close from starting runtime initialization early
- global and named `SPDLOG_LEVEL` apply only to the configured non-stdout sink
- unwritable paths fall back to a null sink; `/dev/full` uses the custom error handler without spdlog stderr diagnostics
- off/null logger flushes and all logger setup errors are exception-contained
- default-logger replacement and weak error-handler lifetime are concurrency-safe for this path
- the production diff is load-bearing and the subprocess tests pin both negative and positive behavior

No blocking findings. Non-blocking follow-ups belong to later #605 parts: the pre-existing dlsym bootstrap fallback, a debug-level rotation re-entry edge, transformer logging, and catastrophic OOM fallback hardening.

yunwei37:
Follow-up independent code review for `ac511ffe8a70878d275c898b2bbd60b636326455`

**Verdict: PASS**

The normal file/console path now directly installs the configured logger and no longer allocates or registers a temporary `bpftime_quiet` logger. The null logger remains failure-only for empty/invalid targets and construction exceptions.

Verified that sink-construction hooks pass through resolved originals while `enable_mock=false`; valid-file startup still captures zero host stdout/stderr and both startup markers; unwritable/default failures remain silent. `/dev/full`, global `SPDLOG_LEVEL`, named `stderr=off`, exception boundaries, and disabled flush behavior are unchanged.

Focused tests: `[logging]` 32 assertions/5 cases and `[allocation]` 43 assertions/7 cases. Production diff is now net +32. No blockers.

yunwei37:
Follow-up external-model review for `ac511ffe8a70878d275c898b2bbd60b636326455`

**External verdict: PASS**

The exact delta from the previously reviewed head is one source line smaller. Source-level review confirmed that configured-sink construction emits no logs before `set_default_logger`; self-interposed open/fopen/close paths are debug-only and pass through resolved originals while mocking is disabled. Failure/empty targets still install an off null sink inside complete exception boundaries.

Global/named `SPDLOG_LEVEL`, `/dev/full` error handling, weak-handler lifetime, and off-level flush behavior remain correct. The change removes the redundant happy-path null logger allocation/registration without weakening the fail-quiet policy. No blockers.

PR #642, opened 2026-08-13
fix(runtime): use no-fault probe memory copies

yunwei37:
## Description

Fix probe read/write fault handling without installing a process-wide SIGSEGV handler.

The existing implementation keeps process-wide handler state in thread-local variables, reads an uninitialized sigaction on repeated calls, and leaves the bpftime handler installed after a helper returns. A SIGSEGV on another thread can therefore enter the wrapper without a recoverable application handler and throw from signal context.

This change:

- uses OS-supported no-fault copies (process_vm_readv/process_vm_writev on Linux and Mach VM operations on macOS);
- leaves application SIGSEGV dispositions untouched;
- preserves checked and unchecked build modes;
- propagates probe-check feature definitions into the runtime test target so guarded fault assertions actually run;
- adds regression coverage for repeated valid access, invalid addresses, cross-thread SIGSEGV delivery, handler replacement, and the default handler.

## Type of change

- [ ] Breaking change

## Validation

- Built bpftime_runtime_tests with both LLVM and uBPF backends.
- Enabled-check focused suites pass: 47 assertions in 5 probe cases and 8 assertions in 2 SIGSEGV cases.
- Disabled-check focused suites pass: 43 assertions in 5 probe cases and 8 assertions in 2 SIGSEGV cases.
- The signal regression cases fail on the previous implementation because it replaces and retains the process-wide handler.

No user-facing API or configuration changes are introduced.

### DIFF
--- a/runtime/src/bpf_helper.cpp
+++ b/runtime/src/bpf_helper.cpp
@@ -11,7 +11,12 @@
 #include <system_error>
 #if __APPLE__
 #include <cstdint>
+#include <mach/mach.h>
+#include <mach/mach_vm.h>
 #include <pthread.h>
+#elif __linux__
+#include <sys/syscall.h>
+#include <sys/uio.h>
 #endif
 #ifdef BPFTIME_BUILD_WITH_LIBBPF
 #include "bpf/bpf.h"
@@ -42,9 +47,6 @@
 #include <bpftime_shm_internal.hpp>
 #include <chrono>
 #include <thread>
-#include <setjmp.h>
-#include <signal.h>
-#include <mutex>
 
 #define PATH_MAX 4096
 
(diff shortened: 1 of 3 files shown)


yunwei37:
Final independent review for exact head c5403e6fa70d1b1b6ad0a2b47e940fb49d3aefe5: **APPROVE**, no blockers.

Reviewed points:
- process_vm_readv/process_vm_writev directions and full-copy checks on Linux;
- mach_vm_read_overwrite direction and copied-size verification on macOS;
- concurrent calls and application SIGSEGV disposition preservation;
- enabled/disabled probe-check builds and test macro propagation;
- ABI declarations, regression coverage, scope, and diff growth.

Non-blocking notes: macOS is source-reviewed because this repository has no macOS CI job; checked Linux helpers add syscall overhead and may return EFAULT under unusually restrictive seccomp policies.

A separate external read-only reviewer also returned APPROVE with no blocking findings.

PR #643, opened 2026-08-13
fix(runtime): implement probe read string semantics

yunwei37:
## Description

Implement the expected string-copy semantics for bpf_probe_read_str.

The helper currently calls strncpy and always returns zero. That loses the copied length, may leave a truncated destination unterminated, and lets invalid pointers terminate the process.

This change:

- returns the copied byte count including the terminating NUL;
- NUL-terminates truncated output and returns the destination size;
- returns 0 for a zero-sized destination, matching Linux no-fault copies;
- returns -EFAULT for inaccessible source or destination memory when probe-read checking is enabled;
- clears the destination after source faults, matching Linux probe-string helpers;
- stops at a terminating NUL on the edge of an accessible page without touching the next protected page;
- adds focused tests for complete copies, truncation, zero length, invalid pointers, page-boundary termination, and faults after a partial copy.

This is stacked on #642 because it reuses the no-fault probe-memory primitives introduced there. It should be retargeted to master after #642 merges.

## Type of change

- [ ] Breaking change

## Validation

- Built bpftime_runtime_tests with both LLVM and uBPF backends.
- Enabled-check focused suites pass: 70 assertions in 6 probe cases and 8 assertions in 2 SIGSEGV cases.
- Disabled-check focused suites pass: 48 assertions in 6 probe cases and 8 assertions in 2 SIGSEGV cases.
- The new string test fails on the previous implementation at the first assertion because it returns 0 instead of 5.
- The final full local runtime suite reached 92/94 passing with 8,561,370 assertions; the two remaining failures are environment-dependent existing tests (kernel tail-call permissions and a relative executable-path check), while all probe tests pass.

No helper IDs, registration entries, public APIs, or configuration defaults are changed.

### DIFF
--- a/runtime/src/bpf_helper.cpp
+++ b/runtime/src/bpf_helper.cpp
@@ -389,12 +389,52 @@ uint64_t bpftime_map_peek_elem_helper(uint64_t map, uint64_t value, uint64_t,
 		.bpf_map_peek_elem((int)map, (void *)value, false);
 }
 
-uint64_t bpf_probe_read_str(uint64_t buf, uint64_t bufsz, uint64_t ptr,
-			    uint64_t, uint64_t)
+#ifdef ENABLE_PROBE_READ_CHECK
+extern "C" void jump_point_read_str();
+#endif
+
+int64_t bpf_probe_read_str(uint64_t buf, uint64_t bufsz, uint64_t ptr,
+			   uint64_t, uint64_t)
 {
-	strncpy((char *)(uintptr_t)buf, (const char *)(uintptr_t)ptr,
-		(size_t)bufsz);
-	return 0;
+	if (bufsz == 0)
+		return -EINVAL;
+
+	int64_t ret = -EFAULT;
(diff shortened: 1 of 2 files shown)


yunwei37:
Final independent review for exact head fdd03b0bfa0b2ea30cad1809c320424fb832d893 against base c5403e6fa70d1b1b6ad0a2b47e940fb49d3aefe5: **APPROVE**, no blockers.

Reviewed points:
- complete-copy, truncation, size-one, and zero-size return semantics;
- invalid source/destination handling and destination clearing;
- terminating NUL at a readable page boundary versus a fault on the next page;
- checked/unchecked build behavior and actual test macro propagation;
- Linux/macOS portability, performance, test scope, and code growth.

Non-blocking notes: checked copies pay per-chunk syscall/Mach-call overhead; a successful string larger than the 256-byte scratch buffer is not explicitly tested.

A separate external read-only reviewer also returned APPROVE with no blocking findings.

Issue #646, opened 2026-08-14
Device-side eBPF: patched kernel not dispatched on GB300 (launch resolves _fini)


EzgiTastan:
## Device-side eBPF: patched kernel never dispatched on GB300 — launch resolves `_fini`

On a **GB300** (Grace, aarch64, CUDA 13.2, driver 595.71.05, kernel 6.14.0-1013), with the `sm_103` PTX + CUDA-13 build fixes applied, the injection pipeline works **end to end through compilation** — but the probe never fires at runtime.

### What works (from the agent debug log)

```
Auto-detected GPU SM arch: sm_103 (compute capability 10.3)
Compiling PTXs with sm_arch sm_103
ptxas info    : Compiling entry function '_Z9vectorAddPKfS0_Pf' for 'sm_103'   (succeeds)
Loaded module: patched.temp.1.sm_75.ptx
Prefilled patched CUDA kernel _Z9vectorAddPKfS0_Pf
```

The patched PTX correctly contains the injected trampoline (`call __probe_func___Z9vectorAddPKfS0_Pf;`) and the eBPF helper calls (`_bpf_helper_ext_*`).

### The bug

Running `example/gpu/cuda-counter` (agent `LD_PRELOAD`, loader `cuda_probe`), the agent logs this on every launch and **no probe events** ever reach the loader:

```
Late attach launch: resolved _fini but no patched CUfunction is registered
```

So at launch the kernel's host stub resolves to `_fini` instead of `_Z9vectorAddPKfS0_Pf`; no patched `CUfunction` matches, and the original (unpatched) kernel runs — the victim's `counter` increments normally, but the eBPF probe is never executed on the GPU.

The patched kernels **are** registered by name at `__cudaRegisterFatBinary` time (`Prefilled patched CUDA kernel _Z9vectorAddPKfS0_Pf` above), so the failure is in the **launch-time** lookup: the host stub pointer → kernel-name resolution mis-resolves on aarch64.

### Environment

- GB300 (Blackwell Ultra, `sm_103`), 2× Grace Neoverse-V2, aarch64
- CUDA 13.2, driver 595.71.05, kernel 6.14.0-1013-nvidia-64k, LLVM 18
- bpftime `master` @ `f311de2` + the `sm_103`/CUDA-13 build fixes

<details><summary>Agent debug log (<code>SPDLOG_LEVEL=debug</code>, key lines)</summary>

```
[...]

PR #647, opened 2026-08-14
feat: support NVIDIA GB300 (sm_103) and CUDA 13

### LINKED ISSUE #645: bpftime rejects injected PTX on sm_103 (GB300) and does not build on CUDA 13
## bpftime rejects injected PTX on `sm_103` (GB300) and does not build on CUDA 13

On an NVIDIA **GB300** (Blackwell Ultra, compute capability 10.3 / `sm_103`), Grace/aarch64, CUDA 13.2, driver 595.71.05, two independent problems stop the CUDA/GPU-attach path from working:

1. **PTX ISA version floor.** `rewrite_ptx_target()` (`attach/nv_attach_impl/nv_attach_utils.cpp`) stamps PTX `.version 8.5` for every `sm_num >= 100`. But `sm_103` was introduced in CUDA 12.9 = **PTX ISA 8.8**, so `ptxas` rejects every injected kernel with a version-too-low error.

2. **CUDA 13 `cuCtxCreate` signature.** CUDA 13 remaps `cuCtxCreate` to the 4-argument `cuCtxCreate_v4` (`CUcontext*, CUctxCreateParams*, unsigned int, CUdevice`). Three call sites (`nv_attach_impl.cpp` ×1, `runtime/src/handler/map_handler.cpp` ×2) use the old 3-argument form and fail to compile on CUDA 13.

Together these mean device-side eBPF can neither be built nor have its PTX compiled on a GB300. A PR fixing both follows (extends #541, which introduced the `rewrite_ptx_target` version ternary).

Environment: GB300 (`sm_103`), 2× Grace Neoverse-V2 aarch64, CUDA 13.2, driver 595.71.05, kernel 6.14.0-1013-nvidia-64k, LLVM 18, bpftime `master` @ `f311de2`.

EzgiTastan:
Fixes #645. Follow-up to #541.

## What this does

Lets bpftime build and compile injected PTX on an NVIDIA GB300 (Blackwell Ultra, `sm_103`) with CUDA 13. Two small fixes:

1. **PTX version for `sm_103`.** `rewrite_ptx_target()` set PTX `.version 8.5` for `sm_num >= 100`. `sm_103` needs `8.8` (CUDA 12.9), so `ptxas` rejected the injected PTX. Added an `sm_103 -> 8.8` case.
2. **CUDA 13 `cuCtxCreate`.** In CUDA 13 `cuCtxCreate` takes 4 args (`cuCtxCreate_v4`). Three call sites used the old 3-arg form and did not compile. Guarded them with `#if CUDA_VERSION >= 13000`.

## How I tested

On a GB300 (Grace/aarch64, CUDA 13.2): it builds, and `example/gpu/cuda-counter` now compiles the injected PTX for `sm_103` (`ptxas ... for 'sm_103'` succeeds) and loads the patched module. Before the fix, `ptxas` rejected the PTX at version 8.5.

## Notes

- On Grace/SBSA the CUDA libs live under `targets/sbsa-linux/`, not the `targets/aarch64-linux/` that `cmake/cuda.cmake` looks for. I used a symlink to build; a `cuda.cmake` change could be a follow-up.
- A separate runtime bug remains: at launch the kernel stub resolves to `_fini`, so the patched kernel is not run. Filed as #646.

Happy to add a small unit test for `rewrite_ptx_target` if you'd like.

## Checklist

- [x] I agree to the CLA (per CONTRIBUTING).
- [x] One root cause; checked all three `cuCtxCreate` call sites.

### DIFF
--- a/attach/nv_attach_impl/nv_attach_impl.cpp
+++ b/attach/nv_attach_impl/nv_attach_impl.cpp
@@ -1216,7 +1216,11 @@ int nv_attach_impl::run_attach_entry_on_gpu(int attach_id, int run_count,
 		CUDA_SAFE_CALL(cuInit(0));
 		CUDA_SAFE_CALL(cuDeviceGet(&cuDevice, 0));
 
+#if CUDA_VERSION >= 13000
+		CUDA_SAFE_CALL(cuCtxCreate(&context, NULL, 0, cuDevice));
+#else
 		CUDA_SAFE_CALL(cuCtxCreate(&context, 0, cuDevice));
+#endif
 		CUDA_SAFE_CALL(cuModuleLoadDataEx(&module, output_elf.data(), 0,
 						  0, 0));
 		// fill data into it
--- a/attach/nv_attach_impl/nv_attach_utils.cpp
+++ b/attach/nv_attach_impl/nv_attach_utils.cpp
@@ -29,7 +29,7 @@ std::string rewrite_ptx_target(std::string ptx,
 
 	// Rewrite .version based on SM arch (sm_120+ needs 8.7, sm_100+ needs 8.5)
 	int sm_num = (sm_arch.size() > 3) ? std::atoi(sm_arch.c_str() + 3) : 0;
-	const char *ptx_ver = (sm_num >= 120) ? "8.7" : (sm_num >= 100) ? "8.5" : nullptr;
+	const char *ptx_ver = (sm_num >= 120) ? "8.7" : (sm_num >= 103) ? "8.8" : (sm_num >= 100) ? "8.5" : nullptr;
 	if (ptx_ver) {
 		auto vpos = ptx.find(".version");
 		if (vpos != std::string::npos) {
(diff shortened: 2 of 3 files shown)


yunwei37 (approved):
LGTM. I reviewed the final head `1eb5c591fe8c33c294224c5aa5053d30b00a84ce`.

What I checked:
- The CUDA 13 `cuCtxCreate` updates match the current CUDA Driver API shape: `cuCtxCreate(CUcontext *, CUctxCreateParams *, unsigned int, CUdevice)`, and `ctxCreateParams` may be NULL for a regular context.
- The PTX target update matches NVIDIA's PTX release history: PTX ISA 8.8 / CUDA 12.9 includes `sm_103`, while the existing `sm_120` and `sm_100` behavior is preserved.
- The PR covers the remaining in-tree CUDA attach `cuCtxCreate` call sites in `attach/nv_attach_impl/nv_attach_impl.cpp` and `runtime/src/handler/map_handler.cpp` for this build path.
- CI is green on this head, including Docker, runtime/example tests, GPU examples, and `test-ptxpass`.
- No existing PR comments or review threads need action.

Nonblocking nit: the comment above `rewrite_ptx_target()` still describes only the old `sm_120+` / `sm_100+` mapping. It would be cleaner to mention `sm_103 -> 8.8`, but I don't think that should block this compatibility fix.

yunwei37:
Thanks！

PR #648, opened 2026-08-14
ci: fetch release tags without submodules

yunwei37:
## Summary
- Fetch only tag refs when resolving the workflow-dispatch release version
- Disable submodule recursion for that tag fetch
- Keep tag-push releases, Docker tag generation, and version bump semantics unchanged

## Why
The manual release dispatch run for v0.9.0 failed while resolving the release version:
https://github.com/eunomia-bpf/bpftime/actions/runs/31842530112

The failing step used `git fetch --tags --force` after a recursive-submodule checkout. In that environment, the broad fetch failed with:

```text
error: expected submodule path 'third_party/libbpf' not to be a symbolic link
```

The resolver only needs local tag refs to avoid reusing an existing release tag, so this narrows the fetch to `refs/tags/*` and explicitly disables submodule recursion.

## Validation
- `git diff --check`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/docker.yml"); puts "yaml ok"'`
- `git fetch --no-recurse-submodules --force origin '+refs/tags/*:refs/tags/*'`
- Local resolver simulation with existing `v0.9.0` tag produced `v0.9.1`

## Release impact
This does not change the already-published v0.9.0 tag or release. It only fixes future manual workflow-dispatch releases.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -77,7 +77,7 @@ jobs:
         id: release-version
         if: "github.event_name == 'workflow_dispatch'"
         run: |
-          git fetch --tags --force
+          git fetch --no-recurse-submodules --force origin '+refs/tags/*:refs/tags/*'
 
           version=$(sed -n 's/^[[:space:]]*VERSION[[:space:]]\+\([0-9]\+\.[0-9]\+\.[0-9]\+\).*/\1/p' CMakeLists.txt | head -n1)
           if [ -z "$version" ]; then


yunwei37:
Review gate summary for `d1a4b692ac4794c04c0b8437ffcd7664beb2512b`:

- Local independent review: PASS. The fetch still provides all `refs/tags/*` needed by the workflow-dispatch resolver, avoids submodule recursion, and does not affect tag-push release behavior, Docker metadata tags, GitHub release publication, or the existing patch-bump loop.
- External independent review: PASS. The diff is one line in `.github/workflows/docker.yml` (`1 insertion, 1 deletion`), with no public API/docs/default/architecture change and no test or fixture bloat. The explicit tag refspec is narrow and sufficient for the failed release-dispatch path.
- Residual validation note: PR CI covers YAML/workflow checks and Docker build for the changed workflow; the actual `workflow_dispatch` publish path can only be proven after this lands on `master`.

PR #650, opened 2026-08-24
Fix hash lookup after deleting colliding keys

yunwei37:
## What

Preserve linear-probing lookup chains after deleting a key from the fixed-size userspace hash map.

The map previously marked a deleted bucket as empty. A lookup for a colliding key stored later in the probe chain would stop at that bucket and incorrectly report the key as missing.

This change introduces an explicit tombstone state:

- lookup stops only at a never-used bucket and skips tombstones;
- delete marks the bucket as deleted instead of empty;
- insert reuses the first tombstone after checking the remaining probe chain for an existing key;
- `is_empty()` continues to mean "not currently occupied" for map iteration.

## Test

Adds a regression section that inserts colliding keys, deletes the first one, verifies the second remains reachable, and verifies the tombstone can be reused by another colliding insertion.

This is intentionally limited to the correctness fix; hash-function and bucket-count performance changes are left out.

### DIFF
--- a/runtime/src/bpf_map/bpftime_hash_map.hpp
+++ b/runtime/src/bpf_map/bpftime_hash_map.hpp
@@ -51,9 +51,9 @@ public:
 // with open addressing and linear probing.
 class bpftime_hash_map {
     private:
-	static constexpr uint32_t bucket_empty = 0;
-	static constexpr uint32_t bucket_filled = 1;
-	static constexpr uint32_t bucket_deleted = 2;
+	static constexpr uint32_t BUCKET_EMPTY = 0;
+	static constexpr uint32_t BUCKET_FILLED = 1;
+	static constexpr uint32_t BUCKET_DELETED = 2;
 
 	inline size_t get_elem_offset(size_t index) const
 	{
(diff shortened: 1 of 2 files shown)


runtime/src/bpf_map/bpftime_hash_map.hpp
```diff
@@ -51,6 +51,10 @@ class bpftime_hasher{
 // with open addressing and linear probing.
 class bpftime_hash_map {
     private:
+	static constexpr uint32_t bucket_empty = 0;
```
Officeyutong: Constants should be in upper case

yunwei37: Updated the three bucket-state constants and every reference to uppercase names in `7f57b36d`. This is an identifier-only follow-up; `git diff --check` passes, and exact-head CI is running.

yunwei37:
A dedicated exact-head code review of `7f57b36d` found no blocking issue. The follow-up commit changes only the three private bucket-state constant names and their references; the full PR tombstone logic and regression coverage remain behaviorally sound, and the size is proportionate to the collision-chain bug. The only human thread is answered and resolved, no Copilot review comment is present, and Officeyutong has approved the exact head.

Exact-head CI is terminal: 159 checks succeeded, 3 were skipped by condition, and none failed or remain pending. The separate independent review gate is the only outstanding process check before this maintenance run can call the PR fully ready.

Officeyutong (approved):

PR #662, opened 2026-08-29
fix: reject truncated daemon program captures

### LINKED ISSUE #655: Daemon can read past captured instruction buffer for programs over 4096 instructions
## Problem

The daemon's kernel-side tracer stores BPF instructions in a fixed buffer sized by `BPF_COMPLEXITY_LIMIT_INSNS` (4096 instructions). For `BPF_PROG_LOAD`, it currently records the original program length in `insns_data.code_len`, but copies only as many instruction bytes as fit in the fixed buffer:

```c
insns_data.code_len = insn_cnt;
read_len = min(sizeof(insns_data.code),
               insn_cnt * sizeof(struct bpf_insn));
bpf_probe_read_user(&insns_data.code, read_len, insns);
```

The userspace driver later trusts `code_len`, resizes the destination vector to that count, and copies `code_len` instructions from the fixed `insn_data.code` array. When `insn_cnt > BPF_COMPLEXITY_LIMIT_INSNS`, this reads beyond the captured buffer and also treats a truncated program as complete.

## Expected behavior

A truncated capture must be rejected before resizing or copying. At minimum, `relocate_bpf_prog_insns()` should return an error when:

```cpp
insn_data.code_len > BPF_COMPLEXITY_LIMIT_INSNS
```

A cleaner protocol could store both the original instruction count and the captured count plus an explicit truncation flag, but the immediate fix should prevent the out-of-bounds read and avoid instantiating an incomplete program.

## Regression coverage

Construct or inject an `insn_data` record whose `code_len` exceeds the fixed capture capacity and verify that:

1. the daemon driver returns a deterministic error;
2. no destination copy is attempted beyond the fixed array;
3. no partial/truncated BPF program is installed into shared memory;
4. a program exactly at the 4096-instruction limit still succeeds.

yunwei37:
## Summary

- reject daemon instruction records whose original count exceeds the fixed 4096-instruction capture buffer
- clear the destination and return `-E2BIG` before resize, copy, relocation, or shared-memory installation
- add focused regressions for an oversized record and the exact 4096-instruction boundary

Closes #655.

## Validation

- focused C++20 guard smoke with `-Wall -Wextra -Werror`
- focused oversized and exact-boundary execution under ASan/UBSan
- `git diff --check`

The full CMake/daemon suite is left to CI because this checkout does not have CMake or initialized submodules.

### DIFF
--- a/daemon/test/test_daemon_driver.cpp
+++ b/daemon/test/test_daemon_driver.cpp
@@ -9,6 +9,7 @@
 #include "bpftime_shm.hpp"
 #include "bpftime_shm_internal.hpp"
 #include "handler/link_handler.hpp"
+#include "user/bpf_prog_insns.hpp"
 #include "user/bpftime_driver.hpp"
 #if !defined(__x86_64__) && defined(_M_X64)
 #error Only supports x86_64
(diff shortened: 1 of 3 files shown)


yunwei37:
Reviewed exact head dbeb27c2c6ec7644ebe58e06baa53010c217cae5 after the full CI matrix reached a terminal state. The capacity check happens before allocation, copying, relocation, or installation; oversized captures return -E2BIG and clear the destination, while the exact 4096-instruction boundary remains accepted. A dedicated source review found no blocking correctness, scope, ABI, or code-growth issue. The +50/-5 production and +31 test delta is proportional to the guard and its two boundary regressions.

Focused GCC 12 strict-warning and ASan/UBSan boundary smokes passed. GitHub GCC 13, daemon build/CLI/tests, Docker build, CodeQL, and GitGuardian checks are green; the two release-only jobs are skipped as expected. The exact head is mergeable, and there are no review or Copilot threads. It now waits on maintainer review.

Officeyutong (approved):

Issue #664, opened 2026-09-03
[FEATURE] Basic support for RISCV


Officeyutong:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

Basic support for RISCV, including frida uprobe support. syscall trace is not expected on RISCV, so does CUDA.
**Describe alternatives you've considered**

**Provide usage examples**

**Additional context**

yunwei37:
This is a useful architecture target. Based on the current tree, the first slice should establish the RISC-V build and runtime baseline before adding uprobe support:

1. Build natively or cross-compile for riscv64 with a clearly selected supported VM path, then cover basic runtime, map, and helper smoke tests.
2. Exclude the syscall-tracing and CUDA-specific components on RISC-V, as requested, instead of reaching their current architecture `#error` paths.
3. Add RISC-V `pt_regs` and Frida Gum register-conversion coverage once a usable Frida backend is available.

The current pinned Frida 16.1.2 release does not provide Linux RISC-V core or Gum devkits, while `cmake/frida.cmake` currently fails for architectures outside its supported list. The Frida portion therefore needs either an upstream RISC-V devkit from a verified Frida release or an explicitly maintained source-build path. A bounded first PR can target the build/runtime smoke path and keep Frida uprobe support as a separately gated follow-up. The main architecture-specific sites to account for are `cmake/frida.cmake`, `attach/frida_uprobe_attach_impl/`, `runtime/src/bpf_map/userspace/ringbuf_map.cpp`, and `runtime/src/handler/perf_event_handler.cpp`.

AI-generated response; a maintainer will review and follow up later

Officeyutong:
Closed by #665

PR #665, opened 2026-09-03
Add trap-based uprobe backend for architectures without Frida

clairjoestar:
Summary

A riscv64-only uprobe attach backend that does not depend on frida-gum. Frida ships no riscv64 devkit (upstream frida-gum PR #1071 is still a draft), so uprobe / uretprobe / uprobe-override / ureplace need an implementation that only relies on the kernel's breakpoint trap.

The new backend lives in attach/trap_uprobe_attach_impl/ and implements the existing base_attach_impl interface, so the runtime and the eBPF program side are unchanged. Other architectures continue using frida untouched.

How it works

1. attach() decodes the first instruction of the target, saves it and overwrites it with ebreak / c.ebreak. When the instruction is naturally aligned the write is a single atomic store. When a 4-byte instruction sits at a 2-byte boundary, a three-phase protocol via c.ebreak ensures every intermediate state is a valid trap (no torn non-trapping word).
2. The SIGTRAP handler looks the faulting pc up in an immutable probe table (atomic pointer swap, no locks in the handler), builds a kernel-style pt_regs from the ucontext_t and runs the callbacks / eBPF programs on the interrupted thread.
3. The original instruction is then either executed out of line from an executable slot followed by a second breakpoint, or emulated when it is pc-relative or a control transfer (auipc / jal / jalr / branches / c.j / c.jr / c.jalr / c.beqz / c.bnez). Because the instruction is never restored in place, no event is lost when many threads hit the probe at once and no free register is required.
4. uretprobes replace the return address (ra) with a breakpoint trampoline and keep the real one on a per-thread shadow stack (embedded in TLS via mmap(MAP_ANONYMOUS), so the first hit on a new thread never allocates through the C library heap; probing malloc itself works). prepare_thread() can be called from normal context to eliminate even the mmap from the first uretprobe hit.
5. Override / replace attaches return to the caller directly from the handler with the requested value.

Host-process contract

- The handler chains to any pre-existing SIGTRAP handler (and puts itself back in front if the host installs one later).

- attach/base_attach_impl/bpftime_pt_regs.hpp: architecture-neutral pt_regs with the riscv64 layout
- cmake/riscv64-toolchain.cmake for cross builds
- cmake/StandardSettings.cmake: compile-time backend selection (set(... CACHE BOOL FORCE))
- Runtime config cleanup: removed use_trap_uprobe_backend field and BPFTIME_UPROBE_BACKEND env var
- Agent: #if/#elif compile-time backend selection instead of runtime branching
- SPDLOG diagnostics on all trap engine error/warning paths
- llvm-jit submodule updated to upstream latest (includes ORC symbol visibility fix)
- .github/workflows/test-riscv64.yml: CI for riscv64 cross-build and qemu-user tests

Tests

- bpftime_trap_uprobe_attach_tests (Catch2): uprobe / uretprobe / override / ureplace, unified base_attach_impl interface, no lost events with 8 threads × 20k calls, attach/detach while the function is hot, recursion, concurrent patching, signal safety, probing malloc from several threads
- runtime/unit-test/attach_with_ebpf/test_attach_uprobe_with_ebpf_trap.cpp: the existing eBPF uprobe test driven through the trap backend
- Verified: riscv64 native on a 32-core rv64gcv host (Release, 200k iterations), qemu-user riscv64

Limitations

- Two signal deliveries per hit; not meant for high-frequency probes.
- The first instruction must be decodable and relocatable; unsupported cases return -ENOTSUP with a diagnostic naming the function.
- Like frida and kernel uretprobes, a hijacked return address is visible to unwinders while the probed function runs.
- Probe sites, out-of-line slots and detached entries are retained for the process lifetime so a racing signal handler can never touch freed memory.
- riscv64 only. Other architectures keep using frida unchanged.

Related to #664


yunwei37 (changes requested):
Thanks for the substantial cross-architecture work. I found three blocking correctness issues in the current head:

- The required `Benchmarks / build-and-run` check fails while linking `bpftime_trap_uprobe_attach_tests`: `__trap_riprel_first` has an undefined reference to `g_trap_global`. Please fix the LTO/release link and restore a green required check.
- `get_uret_stack()` runs on the first uretprobe hit from inside `SIGTRAP`, where it calls `mmap()` and `pthread_setspecific()`. `pthread_setspecific()` is not async-signal-safe, and the exception paths also call spdlog from the handler. If a probe interrupts libc, allocator, or logging state, this can deadlock or corrupt the host, contrary to the stated host-process contract. Please move first-use setup out of the handler or redesign the handler path to use only signal-safe operations, with a new-thread/allocator stress test.
- On riscv64, a 4-byte instruction may be only 2-byte aligned. `write_code()` patches and restores it with two 16-bit stores; during both arming and detaching, another hart can fetch a torn 32-bit instruction. In particular, the low half of the 32-bit `ebreak` is not itself a trapping compressed instruction. Please use a patch/unpatch protocol that cannot expose a torn executable instruction and validate it under concurrent native riscv64 execution.

yunwei37 (changes requested):
Rechecked the new head `bd81a14`. The benchmark link failure from the previous review is fixed and the current checks are green. Two correctness blockers remain in the current source:

- `get_uret_stack()` is still reached from `handle_entry()` inside `SIGTRAP`, and still calls both `mmap()` and `pthread_setspecific()` on first use. Neither is guaranteed async-signal-safe by POSIX. The handler also still reaches spdlog on callback exceptions and other error paths. This does not satisfy the documented host-process contract; please initialize per-thread state outside the handler or otherwise keep the handler path signal-safe, then exercise new threads and allocator/logging interruption under stress.
- `write_code()` still handles a 4-byte RISC-V instruction at a 2-byte boundary with two 16-bit stores. Writing the “trapping half” first does not make the intermediate word safe: the low half of the 32-bit `ebreak` is `0x0073` (a 32-bit instruction prefix), not compressed `c.ebreak` (`0x9002`), and restoration can likewise expose an original low half paired with the trap high half. Another hart can therefore fetch a torn non-trapping instruction. Please use a patch/unpatch protocol whose intermediate states are guaranteed traps (or stop concurrent execution) and validate it under concurrent native riscv64 execution.

yunwei37 (changes requested):
Rechecked exact head `3c7617ffde486ba6376ab1fb9f19f7f6b9f6c3fc`. The three-phase RISC-V patch is a useful improvement, and the normal build/attach/benchmark checks are green, but the prior signal-safety blocker is not closed:

- `get_uret_stack()` still calls `mmap()` from `handle_hit()` on the first uretprobe hit. `prepare_thread()` is only an optional API and has no call site; the new stress tests explicitly exercise lazy allocation inside the handler. POSIX does not guarantee `mmap()` is async-signal-safe. Please make first-hit state available without handler-side allocation, or revise the design so this path cannot occur.
- Removing the pthread destructor also means every short-lived thread that hits a uretprobe retains an mmap allocation until process exit. Please add bounded lifecycle cleanup or another design that does not grow memory/VMA use with thread churn.
- The new three-phase patch test does not force a 4-byte instruction at an address congruent to 2 modulo 4, and the workflow runs qemu-user rather than the previously requested concurrent native riscv64 case. Please make the exact alignment precondition explicit in the test and provide native concurrent validation.
- All trap TLS uses `tls_model("initial-exec")` even though this code is linked into an agent that may be injected after process startup. Please validate late dlopen/injection in a multithreaded host or use a TLS design that does not depend on static-TLS surplus.

The remaining queued GPU jobs are unrelated to these blockers.

Officeyutong (approved):

yunwei37 (changes requested):
Rechecked exact head `cf960cba1d96c382ac31d3f81397efcdb7b3037a`. Embedding the return stack in TLS removes the dedicated per-thread `mmap`, but this head is not ready to merge:

- The signal-safety contract is still unmet. Trap TLS now uses the default dynamic model, while `prepare_thread()` is a no-op. A first access after late injection may resolve/allocate TLS from `SIGTRAP`; the handler also reaches allocating helpers, locks, and newly reintroduced spdlog paths. Please establish all handler state before interception or redesign the handler-reachable path to use only async-signal-safe operations, and test late injection into existing threads.
- The 2-mod-4 RISC-V path is still not exercised: both new tests place the target instruction at `fn + 2` but attach `fn`, whose first instruction is a 2-byte `c.nop`. The workflow is qemu-user only, so there is still no exact-alignment native concurrent patch/restore evidence.
- `write_code()` reports success even when restoring RX permissions fails, potentially leaving a code page RW/RWX. Please propagate that failure and leave the site in a defined safe state.
- The new eBPF trap integration test is not executed by the current RISC-V CI configuration; enabling the full runtime unit tests also hits an unconditional dependency on an unavailable Frida target.
- The broad `vm/llvm-jit` update and the unrelated silent JIT-to-interpreter fallback in `bpftime_prog` should be split out or reduced to the minimum needed for this backend. `git diff --check` also reports the extra EOF blank line in `runtime/src/bpftime_config.cpp`.

yunwei37 (changes requested):
The new exact head `6362703253adfe0f6b5a6c6cec7f529a886be126` closes the earlier signal-handler allocation/TLS-lifecycle, RX-restoration, exact 2-mod-4 patching-test, and RISC-V eBPF-integration concerns. The remaining merge blockers are now narrower:

- The concurrent three-halfword patch protocol still has only qemu-user evidence; it needs an exact-head native RISC-V concurrency run or equivalent maintainer-accepted hardware evidence.
- Exact-head Ubuntu `Build and test runtime` fails with SIGSEGV in `test_attach_filter_with_ebpf.cpp` after 97/98 tests. That job passed at the prior head. Please reproduce the named test alone and the full suite with seed `2702750880`, then isolate the new runtime-test link composition / `--export-dynamic` change from trap-test global state and rerun the runtime matrix.
- `git diff --check` still reports the extra EOF blank line in `runtime/src/bpftime_config.cpp`.
- The PR remains 62 files and about +5,868/-161, including a 12-commit `llvm-jit` update with unrelated atomics, benchmarks, signed-load, pass-ablation, and map/JIT API changes, plus the general JIT fallback. Please split or minimize those unrelated changes, or obtain an explicit maintainer scope decision.

Seven current integration failures inspected are Docker-registry HTTP 500 pulls rather than code failures; those can be retried after the running matrices settle. This review is bound to the exact head above.

yunwei37 (changes requested):
Reviewed exact head `f70c61856dd5ba5b3c08269be19dd5fa28740dd0`. I pushed the one-line EOF cleanup, and the full base-to-head `git diff --check` is now clean. The preceding author update also closes the Ubuntu runtime SIGSEGV: its full 101-test suite passed twice, and this final commit changes only trailing whitespace.

Two substantive gates remain:

- The concurrent RISC-V patching protocol still has only qemu-user evidence. Please provide an exact-head native RISC-V concurrency run or equivalent maintainer-accepted hardware evidence.
- The PR remains 64 files and +5,888/-162, including the unchanged 12-commit `llvm-jit` bump with unrelated work and the general JIT fallback. Please split/minimize those changes or obtain an explicit maintainer scope decision.

Exact-head runtime, RISC-V, benchmark, integration, and GPU matrices are still settling. The currently reported malloc-example failure is infrastructure-only (FridaGum download failed after an external TLS connection reset) and will be retried after the workflow reaches a terminal state. There are no unresolved review or Copilot threads. This review is bound to the exact head above.

yunwei37:
Exact-head CI is now terminal: all 162 reported checks completed successfully, including all 17 jobs in the GPU examples run. This resolves the pending runner evidence from the previous review; it does not change the two remaining merge blockers: native concurrent RISC-V patching evidence and the requested split/minimization (or explicit maintainer acceptance) of the broad 64-file scope.

PR #671, opened 2026-09-04
fix: bounds-check handler fds to prevent OOB and fd leak

### LINKED ISSUE #667: handler_manager: unchecked fd index causes OOB write on pool exhaustion
## What

`handler_manager` uses OS fd values as direct indices into the fixed-size `handlers` array with no bounds check.

- `set_handler(fd, ...)` calls `is_allocated(fd)`, which returns false for any fd past the end, then writes `handlers[fd]`. That write is out of bounds.
- `get_handler(fd)` and `operator[]` read `handlers[fd]` unchecked.
- `open_fake_fd()` returns the OS fd from `open("/dev/null")` and never closes it when the value exceeds the pool size.

The out-of-bounds write corrupts shared memory. The next process that reads the corrupted slot then hits a SIGSEGV, far from the cause.

## Reproduce

- bpftime commit [`5bf24b21af85`](https://github.com/eunomia-bpf/bpftime/commit/5bf24b21af85), still on master [`0834957169da`](https://github.com/eunomia-bpf/bpftime/commit/0834957169da) (2026-08-28)
- Fedora 44, kernel 7.1.6-201.fc44.x86_64

The leak in #668 fills a 128-slot pool in 41 attach/detach cycles. `open_fake_fd` then returns fd 129, used as `handlers[129]`. An in-process harness shows this without Frida or a target process. Rebuilding with the fix returns `-ENOSPC` instead.

Harness and logs: https://github.com/maxgio92/xcover/tree/3fa4a14/patches/bpftime/proofs

## Proposed fix

- `set_handler`: return `-ENOSPC` for out-of-range fds before `is_allocated`.
- `get_handler` / `operator[]`: return a static `unused_handler` sentinel for out-of-range indices.
- `open_fake_fd`: close the fd, set `errno = ENOSPC`, return -1 when the value exceeds `manager->size()`.

Two points for review. The read-path sentinel is process-local memory while in-range handlers are shared memory; no caller computes offsets from the reference today. The sentinel also hides out-of-bounds reads; a `SPDLOG_WARN` there may be better.

Patch: https://github.com/maxgio92/xcover/blob/main/patches/bpftime/0002-fix-handler-manager-oob-and-open-fake-fd-leak.patch

I can open a PR.

maxgio92:
Fixes out-of-bounds handler access and an fd leak when the handler slot pool is exhausted.

handler_manager::set_handler() calls is_allocated(fd), which returns false for any fd >= handlers.size(), then writes handlers[fd] anyway. That out-of-bounds write corrupts memory and SIGSEGVs the caller. get_handler() and operator[] read handlers[fd] with the same missing check.

open_fake_fd() uses the OS-assigned /dev/null fd directly as a handlers[] index. Once the pool is full and the OS hands back a value past the array bounds, the fd leaks and the caller gets an unchecked out-of-bounds index.

set_handler now returns -ENOSPC for out-of-range fds before the is_allocated check. get_handler and operator[] return a static unused_handler sentinel for out-of-range indices. open_fake_fd closes the fd and returns -1 with errno=ENOSPC when the value would exceed manager->size().

Closes #667

### DIFF
--- a/runtime/src/bpftime_shm_internal.cpp
+++ b/runtime/src/bpftime_shm_internal.cpp
@@ -678,6 +678,11 @@ int bpftime_shm::open_fake_fd()
 	while (fd <= 2 && fd >= 0 && --cnt > 0) {
 		fd = dup(fd);
 	}
+	if (fd >= 0 && (std::size_t)fd >= manager->size()) {
+		close(fd);
+		errno = ENOSPC;
+		return -1;
+	}
 	return fd;
 }
 
--- a/runtime/src/bpftime_shm_internal.cpp
+++ b/runtime/src/bpftime_shm_internal.cpp
@@ -679,6 +679,7 @@ int bpftime_shm::open_fake_fd()
 		fd = dup(fd);
 	}
 	if (fd >= 0 && (std::size_t)fd >= manager->size()) {
+		SPDLOG_ERROR("open_fake_fd: fd {} exceeds handler pool size {}", fd, manager->size());
 		close(fd);
 		errno = ENOSPC;
 		return -1;
--- a/runtime/src/handler/handler_manager.cpp
+++ b/runtime/src/handler/handler_manager.cpp
@@ -34,11 +34,15 @@ handler_manager::~handler_manager()
 
 const handler_variant &handler_manager::get_handler(int fd) const
 {
+	static const handler_variant oob_handler = unused_handler{};
+	if (fd < 0 || (std::size_t)fd >= handlers.size()) return oob_handler;
 	return handlers[fd];
 }
 
 const handler_variant &handler_manager::operator[](int idx) const
 {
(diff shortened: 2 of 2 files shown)


runtime/src/bpftime_shm_internal.cpp
```diff
@@ -678,6 +678,11 @@ int bpftime_shm::open_fake_fd()
 	while (fd <= 2 && fd >= 0 && --cnt > 0) {
 		fd = dup(fd);
 	}
+	if (fd >= 0 && (std::size_t)fd >= manager->size()) {
+		close(fd);
+		errno = ENOSPC;
```
Officeyutong, line 684: I think there should be some log printing

maxgio92: Good catch, thanks. Addressed in 1c9c29806a01ed63ccde44b735785e5b3ce796ae

PR #678, opened 2026-09-13
fix: restore CPU affinity on perf output errors

### LINKED ISSUE #677: bpf_perf_event_output leaves thread affinity pinned on error
`bpf_perf_event_output()` can leave the calling thread pinned to a single CPU after returning an error. This changes the host thread's scheduling affinity beyond the lifetime of the helper call.

Reproduced on Linux at `4f785080ad65af2ef2554c2107ccdb094b6fa54a`.

### Reproduction

With initialized bpftime shared memory and a thread allowed to run on at least two CPUs:

1. Save the thread's mask with `sched_getaffinity()`.
2. Call the runtime's `bpf_perf_event_output()` helper with an unknown map FD.
3. Read the affinity mask again after the helper returns `-1` with `errno == ENOENT`.

Observed: the allowed CPU count changes from 16 to 1. Expected: the original affinity mask is restored.

A second path produces the same failure: use a one-entry perf-event array while the thread is allowed to run on at least two nonzero CPUs. The CPU-index lookup fails with `EINVAL`, and the thread remains pinned.

### Cause

In `runtime/src/bpf_helper.cpp`, the helper pins the thread before querying the map type and looking up the CPU's perf-event entry. Both failures return before the final `sched_setaffinity()` cleanup. The initial `sched_getaffinity()` return value is also unchecked.

I have a fix and regression tests prepared and am opening the linked PR immediately. The tests exercise the actual runtime and reproduce both affinity failures against the original helper.

pro-utkarshM:
Fixes #677

`bpf_perf_event_output()` leaves the host thread pinned to one CPU when map-info lookup or CPU-entry lookup fails. Route both errors through common affinity restoration, check that the original mask was saved before pinning, and preserve the output error across cleanup. Report restoration failure when output otherwise succeeded.

Add Linux regression coverage using real runtime maps and affinity syscalls for an unknown map, an out-of-range CPU entry, and successful output. Tests restore their process state and close allocated FDs during cleanup.

Validation:

- Original helper: the new regression fails both affinity assertions (36/38 assertions pass).
- Fixed helper: all 38 regression assertions pass.
- Software perf-event tests: 8,345 assertions pass across 4 test cases.
- Supplemental syscall-failure checks pass for failed affinity capture, failed restoration, and preservation of an existing output error.

Local validation used a focused build of the complete default runtime source list, actual uBPF/VM libraries, and the pinned Catch2 version. Optional Frida, libbpf, and CUDA components were omitted; the full project suite was not run. The regression is included in the existing `bpftime_runtime_tests` target and can be selected with `[affinity]`.


Officeyutong (approved):

yunwei37 (approved):
@/tmp/review-678.md

PR #680, opened 2026-09-13
fix: initialize AOT runner and pass input buffer correctly

### LINKED ISSUE #679: AOT run aborts before execution and passes the wrong input buffer
`bpftime-aot run` does not initialize the local runtime configuration before constructing its program. Once that is initialized, it also passes the `std::vector` object itself as program memory and reports load/execution failures as successful CLI exits.

Reproduced on Linux at `4f785080ad65af2ef2554c2107ccdb094b6fa54a` with the pinned LLVM backend and LLVM 20.1.8.

### Reproduction

Compile a small native object exporting the entry point accepted by the AOT loader:

```c
#include <stddef.h>
#include <stdint.h>

uint64_t bpf_main(void *memory, size_t size)
{
    const uint8_t *bytes = memory;
    if (size == 0)
        return 0;
    return ((uint64_t)size << 32) | bytes[0] |
           ((uint64_t)bytes[size - 1] << 8);
}
```

```sh
cc -O2 -fPIC -c fixture.c -o fixture.o
printf '\052\000\000\000' > memory.bin
bpftime-aot run fixture.o memory.bin
```

Expected: successful execution with `Output: 17179869226`, reflecting four input bytes and their contents.

Actual: the unmodified CLI aborts with `VM name cannot be empty`, including when `BPFTIME_VM_NAME=llvm` is set. Its shared-memory holder/local configuration has not been initialized. With only local initialization added, the program receives a length of 24 on this machine instead of 4, because `run_ebpf_program()` passes `&mem, sizeof(mem)` rather than the vector's data and length. Omitted and empty input similarly appear nonempty.

There are also two related failure-handling problems in the same function: the result of `load_aot_object()` is ignored, and the execution-error branch returns CLI status zero. With local initialization alone, malformed objects and objects missing `bpf_main` incorrectly exit zero.

I have a fix and nine real-CLI regression cases prepared and am opening the linked PR immediately. A program returning `UINT64_MAX` remains a successful execution; that program result must not be confused with a runtime failure.

pro-utkarshM:
Fixes #679

`bpftime-aot run` aborts before execution because its local runtime configuration is uninitialized. After initialization, it supplies vector bookkeeping instead of the input file's bytes and reports failed AOT loads/execution as successful CLI exits.

Initialize the existing runtime holder in `SHM_NO_CREATE` mode and set its configuration before constructing the program. Pass `mem.data(), mem.size()`, stop immediately on AOT-load failure, and return status 1 for execution failure. Successful program return values remain independent of the CLI exit status. Initialization creates no shared-memory segment.

Add a standard-library Python regression that compiles native fixtures and exercises the real CLI/LLVM loader. It covers short/long input, omitted/empty input, `UINT64_MAX` as a successful program result, malformed/missing objects, missing entry points, and missing input files. Run it in the existing Ubuntu and Fedora tools workflow:

```sh
python3 tools/aot/test_run.py build/tools/aot/bpftime-aot cc
```

Validation:

- Unmodified CLI: seven cases abort on missing runtime configuration; the two missing-file cases pass.
- Initialization alone: four input cases and two load-status cases fail, isolating the remaining defects.
- Complete fix: all nine cases pass.
- Python syntax, workflow YAML parsing, and `git diff --check` pass.

Local validation used an isolated build of the actual CLI, default runtime sources, and pinned LLVM backend with LLVM 20.1.8. The complete root CMake configuration and CI matrix were not run locally. The post-load execution-status guard is defensive: successful AOT loading takes the current runtime's direct JIT path, so the test does not introduce a mocked VM seam to force that branch.


yunwei37 (approved):
@/tmp/review-680.md

Officeyutong (approved):

PR #687, opened 2026-09-18
fix: close temporary kernel object FDs in daemon

### LINKED ISSUE #686: Daemon leaks temporary kernel program and map FDs
The daemon leaks temporary kernel-object descriptors acquired by `bpf_map_get_fd_by_id()` and `bpf_prog_get_fd_by_id()`.

At upstream `edcc8ec`, in `daemon/user/bpftime_driver.cpp`:

- `bpftime_maps_create_server()` never closes its temporary map FD, including on successful creation and when the object already exists.
- `bpftime_progs_create_server()` closes its temporary program FD only on final success. Metadata lookup, relocation, shared-memory insertion, dependent-map lookup, and already-present returns leave it open.

Repeated calls can exhaust the daemon's descriptor limit and keep kernel BPF objects referenced longer than intended. The runtime's kernel-backed map implementations acquire their own descriptors when needed; they do not take ownership of these temporary descriptors.

I reproduced this with a regression test linked against the real daemon and runtime, using libbpf wrappers that return real `/dev/null` descriptors and controlled metadata/errors. Against unchanged upstream, the test reports 27 failed assertions: affected returns leave the acquired descriptor valid according to `fcntl(F_GETFD)`. After adding scope cleanup, the full daemon suite passes 124 assertions across 5 test cases.

Expected behavior: close each successfully acquired temporary descriptor exactly once, keep the program descriptor open through dependent-map enumeration, and preserve the original error across cleanup.

I have prepared the fix and regression coverage on `pro-utkarshM:fix/daemon-kernel-fd-leaks` and am opening the linked PR immediately.

pro-utkarshM:
The daemon leaves temporary map FDs open on every return and program FDs open on early returns. Repeated imports can exhaust descriptors and retain kernel BPF objects unnecessarily.

Fixes #686.

Add Boost.ScopeExit guards immediately after successful acquisition in both driver functions, preserving `errno`. Keep the program FD valid through dependent-map enumeration and remove its old success-only close. Production change: 13 added lines, one removed line.

Regression coverage uses the real daemon/runtime with test-only libbpf wrappers and real descriptors, without BPF privileges. Twelve scenarios cover success, existing objects, failed acquisition, metadata lookup, insertion, relocation, and dependent-map enumeration. Checks verify closure exactly once, descriptor liveness until its final use, and error preservation. The tests run in the existing daemon test target and CI job.

Validation:

- Unchanged upstream plus the regression test: 27 expected failed assertions from leaked FDs.
- Fixed daemon suite: **124 assertions passed across 5 test cases**.
- Built both `bpftime_daemon_tests` and `bpftime_daemon` with the uBPF backend.
- `git diff --check` and formatting check for the new test passed.

Local environment: GCC 16, Boost 1.87 headers. Building the existing bundled bpftool required a local `-Wno-error=discarded-qualifiers` override and `feature-clang-bpf-co-re=0` to skip its optional profiler skeletons on this newer host. Those overrides are not repository changes; the daemon's BPF object and skeleton were built. Validation did not load BPF objects into the kernel.

Test command:

```sh
BPFTIME_GLOBAL_SHM_NAME=bpftime_daemon_fd_validation \
  ./build/daemon/test/bpftime_daemon_tests
```


yunwei37 (commented):
Verified the fix and the regression coverage at head `77ec23f765`. The leak is real and the change is correct.

**Both leaks confirmed in the upstream source at `edcc8ec`.** `bpftime_maps_create_server()` acquired `map_fd` from `bpf_map_get_fd_by_id()` and never closed it on any path, including the already-exists return and the success return. `bpftime_progs_create_server()` closed `fd` only at the very end, so the metadata-lookup, relocate, shm-insert, related-map, and already-present returns all leaked it.

**The fix is correct on the path that matters most.** The one place that must keep `fd` alive is the dependent-map enumeration:

```c
res = check_and_create_prog_related_maps(fd, &info);
if (res < 0) { ... return -1; }
```

`check_and_create_prog_related_maps()` re-reads program info through that same `fd`, so a naive close before it would have broken related-map discovery. Replacing the trailing `close(fd)` with a `BOOST_SCOPE_EXIT_ALL` guard closes on every return, including the error paths, while leaving `fd` valid for that call. Saving and restoring `errno` around the close is the right detail, since these paths return `-1` for the caller to inspect.

**The regression test is genuine and it ran in CI.** `build-gcc13.yml` configures `-DBUILD_BPFTIME_DAEMON=ON`, builds the `bpftime_daemon_tests` target, and runs it, so this suite is exercised for this PR rather than being dead code. The log for this head shows `test_daemon_fd.cpp` compiling and then:

```
All tests passed (124 assertions in 5 test cases)
```

which matches the count in the report. The test wraps `close` and the libbpf fd-acquisition calls, so it asserts both that a descriptor was closed exactly once and that it was still valid beforehand — that is a real behavioural check rather than a mock echo.

Two observations for the record, neither blocking:

- The `__real_*` wrappers return descriptors opened on `/dev/null`, so what is verified is descriptor lifetime and accounting in the driver, not kernel-object reference behaviour. That is the right scope for this bug and worth stating in the PR body so the coverage is not over-read.
- `master` requires one approving review and has no required status checks. The review is what unblocks this.

Thanks for the clear reproduction and the linked test — the failure-path coverage here (error before acquisition, error after acquisition, already-present, success) is more thorough than the fix strictly needed.

yunwei37 (approved):
Approving: the defect is confirmed in the source, the fix is correct on the path that could have regressed, and the regression test is exercised by CI on this head.

Evidence:
- Both leaks reproduce from reading upstream `edcc8ec`: the map descriptor is never closed on any path, and the program descriptor is closed only on final success, so five other returns leak it.
- The fix keeps `fd` valid through `check_and_create_prog_related_maps(fd, &info)`, which re-reads program info through that descriptor, then releases it via a scope-exit guard on every return. `errno` is preserved across the close.
- `build-gcc13.yml` configures `-DBUILD_BPFTIME_DAEMON=ON` and runs `bpftime_daemon_tests`, so the new test is not dead code. The log on this head shows `test_daemon_fd.cpp` compiling and the suite reporting 124 assertions across 5 test cases, all passing.
- `Build GCC 13` and `Build and push docker image` are green at this head.

This unblocks the review requirement; per repository policy the merge decision remains with the maintainer.

PR #689, opened 2026-09-18
fix: return epoll readiness without duplicates and restore signals

### LINKED ISSUE #688: epoll_wait skips zero-timeout readiness, leaks signal masks, and duplicates events
`bpftime_epoll_wait()` in `runtime/src/bpftime_shm.cpp` has two readiness-loop defects at upstream `edcc8ec`:

1. With `timeout == 0`, it returns before examining registered sources and before restoring the signal mask. A ready source is missed, and initially unblocked SIGINT/SIGTERM remain blocked in the calling thread.
2. After finding events, it sleeps and rescans the same registrations until `maxevents` is filled or the timeout expires. One unread ready registration can therefore appear multiple times in a single result batch.

Reproduced using the real runtime with one ring-buffer registration, published data, registration data `77`, and an output capacity of four:

```text
timeout=0: returned 0 events; SIGINT remained blocked
timeout=10: returned 4 events; data=77,77,77,77
```

Expected: a zero-timeout call polls current readiness without waiting, each ready registration appears once in the returned batch, and the caller's signal mask is restored. This follows the documented epoll_wait contract: https://man7.org/linux/man-pages/man2/epoll_wait.2.html

The fix can reuse the existing cleanup path: remove the pre-scan zero-timeout return, then leave the loop after its first ready batch or after one scan for timeout zero.

I have prepared a fix and regression tests covering real ring buffers and software perf events, zero/finite/indefinite-ready waits, duplicate prevention, output bounds, repeated polling, and signal masks. I am opening the linked PR immediately.

pro-utkarshM:
Zero-timeout `bpftime_epoll_wait()` calls currently skip ready sources and leave SIGINT/SIGTERM blocked. Positive or indefinite waits can return the same ready registration repeatedly while filling the output buffer.

Fixes #688.

Remove the early zero-timeout return and exit after a readiness scan when events were found or timeout is zero. Both paths then use the existing signal-mask restoration. The production diff is **2 added lines and 6 removed lines**; empty positive/infinite waiting behavior is unchanged.

Add `[epoll]` regression coverage to the existing runtime test target. Forty scenarios use real ring buffers/software perf events and two starting signal masks; each scenario polls twice to verify unread data remains ready. Coverage includes empty and ready sources, zero/finite/indefinite-ready waits, distinct registrations, and output capacity. Tests restore signals, environment, shared memory, and descriptors after failures too. No new dependencies or workflow changes.

Validation (local uBPF build):

- Before the production fix: **172 failing assertions** demonstrating both defects.
- After: `[epoll]` passes **6,558 assertions**; `[perf_event]` passes **8,345 assertions across 4 test cases**.
- Runtime test target builds; whitespace and new-test formatting checks pass.
- Full-suite attempt: **97 test cases passed, 7 failed**, ending in a Frida/GLib crash. The same unrelated failure categories were reproduced before rebuilding the fix: five tests explicitly require the unavailable LLVM backend, one kernel tailcall test cannot create its BPF map, and the attach-replace test crashes in Frida symbol lookup. This is not a claim that the full local suite passes. Existing runtime CI provides LLVM-backed validation.

Focused commands:

```sh
CPLUS_INCLUDE_PATH=/tmp/boost_1_87_0 \
  cmake --build build --target bpftime_runtime_tests -j4
./build/runtime/unit-test/bpftime_runtime_tests '[epoll]'
./build/runtime/unit-test/bpftime_runtime_tests '[perf_event]'
```


yunwei37 (commented):
Verified at head `a7b36bccab`. Both defects are real and the fix addresses them.

**Confirmed in the upstream source at `edcc8ec`.** The readiness loop returned before doing any work:

```c
while (next_id < max_evt) {
        ...
        if (timeout == 0)
                return 0;
```

That early return skips the scan entirely, so a ready source is missed, and because it bypasses the `sigprocmask(SIG_SETMASK, &orig_sigset, nullptr)` that only runs after the loop, the SIGINT/SIGTERM the function blocked stay blocked in the calling thread. The duplicate half comes from the `sleep_for(1ms)` followed by a rescan of the same registrations: an unread ready source is re-appended on each pass until `max_evt` is reached.

**The fix is correct and reuses the existing cleanup path.** Removing the zero-timeout return lets a zero timeout fall through to one scan, and changing the post-scan branch to `if (next_id > 0 || timeout == 0) break;` exits after the first ready batch instead of sleeping and rescanning. Every exit now reaches the mask restore, since breaking out of the loop leads to it. The `timeout == 0` arm also means a poll that finds nothing returns promptly rather than entering the signal-wait path.

**The regression coverage is real, exercised, and well chosen.** `test_epoll.cpp` is registered in `runtime/unit-test/CMakeLists.txt`, and `test-runtime.yml` triggers on `runtime/**` and builds with `BPFTIME_ENABLE_UNIT_TESTING=YES`, so it is not dead code. The run on this head shows it compiling with 90% line coverage and the suite reporting:

```
All tests passed (8586401 assertions in 104 test cases)
```

The case matrix targets exactly the two defects rather than only the happy path — `two ready polls` with one registration and a capacity of four asserting `expected == 2` is what catches the duplicate bug, and the `timeout == 0` cases are what catch the missed-readiness bug. It also checks `count <= capacity`, that waiting does not consume the data, and that the signal mask is restored for both SIGINT/SIGTERM.

Two small notes, neither blocking:

- The fix changes observable behaviour for a genuine outlier: a caller that previously relied on `timeout == 0` returning `0` unconditionally now gets real readiness. That is the documented `epoll_wait` contract and the point of the change, so it is the right direction, but it is worth a line in the PR description since it is a behaviour change rather than only a bug fix.
- The `sleep_for(1ms)` remains for the finite-timeout path, where it still re-scans after sleeping; with the break added, a batch now ends at the first ready pass, so the sleep only elapses when nothing is ready yet. That matches the intent of not spinning, and I do not see a case it breaks.

Thanks for the precise reproducer — the concrete `timeout=0: returned 0 events; SIGINT remained blocked` line made this quick to confirm against the source.

yunwei37 (approved):
Approving: both defects are confirmed in the source, the fix is minimal and reaches every exit path, and the new test is built and run by CI on this head.

- Upstream `edcc8ec` returns `0` from inside the readiness loop when `timeout == 0`, which skips the scan and bypasses the `sigprocmask(SIG_SETMASK, ...)` restore that only runs after the loop, leaving SIGINT/SIGTERM blocked in the caller. The `sleep_for(1ms)` plus rescan re-appends an unread ready registration until `max_evt` fills.
- The fix removes the early return and exits the loop on `next_id > 0 || timeout == 0`, so a zero timeout performs one scan and every exit reaches the mask restore. It reuses the existing cleanup path rather than adding a second one.
- `test-runtime.yml` triggers on `runtime/**` and builds with `BPFTIME_ENABLE_UNIT_TESTING=YES`; the run on this head compiles `test_epoll.cpp` with 90% line coverage and reports `All tests passed (8586401 assertions in 104 test cases)`. The case matrix exercises both defects directly: `two ready polls` with one registration and capacity four asserting two events catches the duplicates, and the zero-timeout cases catch the missed readiness and the signal mask.
- All other workflows are green at this head.

This satisfies the review requirement. Per repository policy the merge decision remains with the maintainer.

PR #691, opened 2026-09-21
fix: report current UID and GID in identity helper

### LINKED ISSUE #690: bpf_get_current_uid_gid returns cached GID in both halves
`bpf_get_current_uid_gid()` in `runtime/src/bpf_helper.cpp` caches `getgid()` in a static `int` and returns that value in both halves of the result:

```cpp
static int gid = getgid();
return (((uint64_t)gid) << 32) | gid;
```

The [helper contract](https://man7.org/linux/man-pages/man7/bpf-helpers.7.html) requires the current UID in the lower 32 bits and current GID in the upper 32 bits. Consequently, different UID/GID values produce an incorrect UID, and credential changes after the first call are not reflected. The signed intermediate also risks sign extension for high-bit IDs.

Reproduced against upstream `edcc8ec1c0d9cff7a241ad7b7467117a81887878` by linking the actual runtime helper with Linux linker wrappers for `getuid()` and `getgid()` (without changing real process credentials):

| Query results | Expected helper UID/GID | Observed UID/GID |
|---|---|---|
| UID 1001, GID 2002 | 1001 / 2002 | 2002 / 2002 |
| UID 3003, GID 2002 | 3003 / 2002 | 2002 / 2002 |
| UID 3003, GID 4004 | 3003 / 4004 | 2002 / 2002 |

The fix should query both current credentials on every call and pack them using unsigned conversions. #685 addresses the separate PID/TID cache after fork and explicitly leaves this helper unchanged.

I have the fix and a focused regression ready and am opening the linked PR immediately. The regression covers distinct IDs, independent UID/GID changes, zero and high-bit IDs, and real-credential fallback.

pro-utkarshM:
When UID and GID differ, `bpf_get_current_uid_gid()` returns GID as UID. It also caches the first GID forever, so later credential changes are invisible. Query `getuid()` and `getgid()` on each call and pack the current values into their specified unsigned halves.

Closes #690.

Add a Linux Catch2 regression using linker wrappers around the credential queries. It calls the actual runtime helper with distinct IDs, independent UID/GID changes, zero and high-bit IDs, then checks real credentials after the scoped override. Outside that thread-local override, the wrappers forward to libc; no privileges or process credential changes are needed. The wrapper options apply only to the Linux test executable.

This deliberately adds two credential queries per helper call instead of retaining stale cached values. No public API or helper registration changes. The PID/TID fix in #685 explicitly left UID/GID unchanged.

### Validation

- Built `bpftime_runtime_tests` locally with the existing uBPF configuration.
- New `[uid_gid]` regression: all 8 assertions failed before the production fix; all 8 pass afterward.
- Existing PID/TID fork regression: all 6 assertions pass.
- `git diff --check` and the new test's clang-format check pass.
- Full runtime suite was run on unmodified upstream and the patched tree: both report 96 passing cases and the same 7 failing locations, ending in a Frida attachment SIGSEGV. Five failures request the unavailable LLVM backend, one is kernel map creation, and one is the Frida crash. The full local suite is therefore not green; the new regression was verified separately. LLVM-backed coverage remains for upstream CI.

Commands used:

```sh
CPLUS_INCLUDE_PATH=/tmp/boost_1_87_0 \
  cmake --build build --target bpftime_runtime_tests -j4
./build/runtime/unit-test/bpftime_runtime_tests '[uid_gid]'
./build/runtime/unit-test/bpftime_runtime_tests \
  'bpf_get_current_pid_tgid reports the calling process after fork'
./build/runtime/unit-test/bpftime_runtime_tests
```


yunwei37 (approved):
Verified the fix against the source at upstream `edcc8ec`, and the diagnosis is correct.

Before this change `bpf_get_current_uid_gid` returned `(((uint64_t)gid) << 32) | gid`, so the lower half carried the GID, and the `static` initializer froze both halves at first use. The current helper queries both credentials per call and packs the GID into the upper half and the UID into the lower half with unsigned conversions, which matches the kernel helper contract and also removes the sign-extension risk on high-bit IDs.

I checked the regression actually runs rather than only that a check is green: `runtime/unit-test/test_get_current_uid_gid.cpp` is compiled into `bpftime_runtime_tests`, and the `test-runtime` workflow builds that target and executes it on both the ubuntu-2204 and fedora-39 containers. The log shows the file built and the suite finishing with `All tests passed (8570351 assertions in 104 test cases)` at this head, with the new file exercised rather than skipped. The `--wrap=getuid`/`--wrap=getgid` approach correctly leaves real credentials untouched for the rest of the suite, and the scoped restore means other tests still observe libc.

Nothing to change from my side.

PR #693, opened 2026-09-21
Update llvmbpf submodule to trim linked LLVM backends

clairjoestar:
Pulls in eunomia-bpf/llvmbpf#52, which links and registers only the host LLVM backend (LLVM_NATIVE_ARCH) plus NVPTX and SPIR-V instead of every backend listed in LLVM_TARGETS_TO_BUILD. With distro LLVM 17 on x86-64 this shrinks libbpftime-agent.so from 140 MB to 79 MB and libbpftime-syscall-server.so from 117 MB to 56 MB (Release build), with .rodata going from 46 MB to 11 MB.

Also picks up the BTF line info debug metadata work (#50) that landed in between.


yunwei37 (approved):
Verified the bump against the llvmbpf repository and it is exactly what the description says.

The gitlink moves `8fc0ceee` → `87bac39a`, which is a 3-commit range: `ef6a10a5` (BTF line-info debug metadata, already merged into bpftime), `c3f92f5a` (its regression test), and `87bac39a` (`llvmbpf#52`, host-backend-only linking). `llvmbpf#52` was merged to llvmbpf's `main` by a maintainer, so the recorded commit is the reviewed upstream state rather than an unmerged fork branch, and the two extra commits are the already-landed BTF work rather than unrelated changes.

The linked-backend reduction is consistent with the source: `LLVM_COMPONENTS` drops `${LLVM_TARGETS_TO_BUILD}` — which distro LLVM configs expand to every backend — for `${LLVM_NATIVE_ARCH}` plus NVPTX, and `InitializeAllTarget*` calls are replaced by the VM's own registration. Keeping NVPTX preserves the GPU `generate_ptx()` path, and SPIR-V remains conditional on its option, so no existing target loses support on the architectures the project builds for.

At this head all 159 checks report success with no failures across `Build and test runtime`, `Build and test tools`, `Build and Attach Tests`, `Build and Test VM`, `Build and run bpftrace test`, and `Build and test PTX pass executables`. Only the self-hosted `test-gpu-examples` fan-out is still queued, which is runner capacity rather than a result of this change.

LGTM from my side.

